PR #5, opened 2022-06-27
Writing a github action to deploy to staging

goldblatt:
writes a github action that would deploy to staging every 12 hours


goldblatt:
needs the secrets!

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,27 @@
+name: Deploy
+on:
+  push:
```
goldblatt, line 3: if you want to submit this please remove this since it will deploy to staging on any push to the repo!

azizova-leyla (approved):

PR #6, opened 2022-08-09
Update civiform_config.sh to match current deployment. 

nbeloglazov:


azizova-leyla (changes requested):
Please remove variables that are not relevant for this deployment, e.g. azure stuff.

azizova-leyla (approved):

PR #10, opened 2022-08-11
Add prober to run after each deploy

azizova-leyla:


bin/run-prober
```diff
@@ -0,0 +1,11 @@
+#! /usr/bin/env bash
+
+# DOC: Run the browser test suite in Docker against the staging environment.
+docker pull civiform/civiform-browser-test:latest
+
+docker run \
+  -e BASE_URL="https://staging-aws.civiform.dev" \
+  -e TEST_USER_LOGIN="${TEST_USER_LOGIN}" \
+  -e TEST_USER_PASSWORD="${TEST_USER_PASSWORD}" \
+  civiform/civiform-browser-test:latest \
```
nbeloglazov, line 10: Sooo, it will run the tests which are whatever in the "latest" browser-test image. In the previous action you deployed civiform with `--tag={LATEST_SNAPSHOT_TAG}`. Will they both (civiform-browser-test and civiform) images be at the same commit? Otherwise it's possible that tests will be out-of-sync with the version that's running on staging.

azizova-leyla: per offline conversation - we don't version browser test image right now.

nbeloglazov (approved):

PR #15, opened 2022-08-18
Remove bin/bash, it's not in the entrypoint

azizova-leyla:
https://github.com/civiform/civiform/pull/3187 added bin/bash as an entrypoint. removing extra from this file.

### DIFF
--- a/bin/run-prober
+++ b/bin/run-prober
@@ -5,6 +5,6 @@ docker pull civiform/civiform-browser-test:latest
 
 docker run \
   -e BASE_URL="https://staging-aws.civiform.dev" \
-  --init \
   civiform/civiform-browser-test:latest \
-  --entrypoint /bin/bash -c "yarn install && yarn probe"
+  -c "yarn install && yarn probe" \
+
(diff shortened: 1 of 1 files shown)


bin/run-prober
```diff
@@ -6,4 +6,4 @@ docker pull civiform/civiform-browser-test:latest
 docker run \
   -e BASE_URL="https://staging-aws.civiform.dev" \
   civiform/civiform-browser-test:latest \
-  /bin/bash -c "yarn install && yarn probe"
+  -c "yarn install && yarn probe"
```
nbeloglazov, line 9: I don't fully understand now what it does. Who receives "-c" option? Is it a default script which is part of civiform-browser-test?

azizova-leyla: it is pretty confusing, but that's how docker entrypoints work. The entry point is /bin/bash and defined in https://github.com/civiform/civiform/blob/615e7008b1c3e4273ab34556a1e6b692d5b7e5ca/browser-test/playwright.Dockerfile#L27.

nbeloglazov: I see. Can you add coment saying that we are relying on entrypoint from the docker which is bash.

nbeloglazov (approved):

PR #27, opened 2022-09-20
add ability to run terraform from run.py

azizova-leyla:
This now works:
cloud/shared/bin/run.py --command="terraform init"
cloud/shared/bin/run.py --command="terraform force-unlock LOCKID"

### DIFF
--- a/cloud/shared/bin/deploy.py
+++ b/cloud/shared/bin/deploy.py
@@ -6,7 +6,7 @@ from typing import List
 from cloud.shared.bin.lib.config_loader import ConfigLoader
 
 
-def run(config: ConfigLoader, params: List[str] = None) :
+def run(config: ConfigLoader, params: List[str] = None):
     deploy_file_py = os.path.join(
         'cloud', config.get_cloud_provider(), 'bin', 'deploy.py')
     # TODO(#2741): remove the fork after Azure scripts are in python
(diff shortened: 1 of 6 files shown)


cloud/shared/bin/deploy.py
```diff
@@ -5,7 +5,7 @@
 from cloud.shared.bin.lib.config_loader import ConfigLoader
 
 
-def run(config: ConfigLoader):
+def run(config: ConfigLoader, params=None):
```
nbeloglazov, line 8: Given that run() now always called with 2 arguments and the second argument is always List[str], make it `run(config: ConfigLoader, params: List[str])`?

nbeloglazov:
How will calling terraform look in the `civiform-deploy` repo?

azizova-leyla:
I need to update civiform-deploy to make it work.
I will likely change` bin/run` to accept --command as argument that will be passed in here. Otherwise it gets tricky with multiword commands.

cloud/shared/bin/destroy.py
```diff
@@ -1,13 +1,14 @@
...
 """
+from typing import List
 
 from cloud.shared.bin.lib import terraform
 from cloud.shared.bin.lib.config_loader import ConfigLoader
 from cloud.shared.bin.lib.setup_class_loader import get_config_specific_destroy
 
 
-def run(config: ConfigLoader):
+def run(config: ConfigLoader, params: List[str] = None):
```
nbeloglazov, line 11: nit: you don't need to provide = None given that second parameter is always passed.

nbeloglazov (approved):

PR #41, opened 2022-09-22
Run pytest on every pr.

azizova-leyla:
Fixes https://github.com/civiform/civiform/issues/2888


.github/workflows/run_pytest.yaml
```diff
@@ -0,0 +1,27 @@
+name: PyTest
+
+on:
+  # code review question: this commits directly to PR branches.
+  # that seems like what we're looking for, but we could also format only
+  # committed code.
```
nbeloglazov, line 6: I don't understand that. The action is just running test. Why are are talking about "commiting" and "formatting"?

cloud/shared/variable_definitions.json
```diff
@@ -41,6 +41,12 @@
     "tfvar": true,
     "type": "string"
   },
+  "ADFS_ADMIN_GROUP": {
```
nbeloglazov, line 44: Was this variable lost in some previous PR?

.github/workflows/run_pytest.yaml
azizova-leyla: Copied it from somewhere else. Not applicable here.

cloud/shared/variable_definitions.json
azizova-leyla: it was in both aws and azure configs, and currently that's not allowed. I'm moving it to shared config.
cloud/shared/validate_variable_definitions_test.py is checking for that and it wasn't running before this change.

nbeloglazov (approved):

PR #56, opened 2022-10-04
Pipe `ALLOW_CIVIFORM_ADMIN_ACCESS_PROGRAMS` in AWS deployment

nbeloglazov:
That variable controls whether CiviForm can see applications for a given program. Feature already implemented in CiviForm but we don't propagated in AWS deployment scripts. So entities who use our deploy scripts can't set it.


avrittrohwer (approved):
Should we add this to the the civiform_config.example.sh file as well?

nbeloglazov:
I don't think so. We don't specify every single option in the example file. As it already has a lot of variables and getting overwhelming. I think we should provide a doc that lists all possible variables with comments of what they do. And keep .example.sh small-ish. That might be a good thing to do while working on migrating configuration files to yaml file. Where we can group variables better.

PR #77, opened 2022-11-01
Add support for pgadmin container.

avrittrohwer:
Terraform implementation for https://github.com/civiform/civiform/issues/2563.

New resources are disabled by default by using count = 0 on the module.

This will be followed up with an additional PR that sets count = 1 via a python command wrapper.

Tested:
1. Set count to 1, built a new docker image, ran `USE_DOCKER=true bin/deploy --tag=latest` in my config repo.
2. Terraform successfully deployed the pgadmin resources.
3. I navigated to https://avritt-dev.civiform.dev:4433 and successfully logged in using the created username and password.
4. I opened the CiviForm db group which prompted for the database password.  I successfully authenticated to the postgres database.
5. I successfully queried the postgres database with `SELECT * FROM questions` which returned the 2 default questions.
6. Set the count to 0, built a new docker image, ran `USE_DOCKER=true bin/deploy --tag=latest` in my config repo.
7. Terraform successfully destroyed the pgadmin resources.
8. I manually checked the lb listener rules, target groups, ecs services, ecs tasks, and secrets in the AWS UI to ensure the resources were properly deleted.

PR #79, opened 2022-11-04
Add pgadmin command

avrittrohwer:
Terraform code added in https://github.com/civiform/cloud-deploy-infra/pull/77.

Fixes https://github.com/civiform/civiform/issues/2563.

Example CLI output:

```
$ bin/run --tag=latest
Using config file civiform_config.sh
Enter command to run : pgadmin
Using civiform/civiform-cloud-deployment:latest
Running: pgadmin --tag=latest --config=civiform_config.sh
Do you want to proceed? [y/N] y
Getting config from civiform_config.sh
Writing TF Vars file


REQUIRED: input a comma-separated list of IPv4 CIDR blocks that should have access to the pgadmin service:
To allow a single IP address, use a subnet mask of 32. For example: '192.0.0.1/32'
> 75.172.108.49/32
Parsed list: ["75.172.108.49/32"]

Accept CIDR list? (anything entered other than 'y' will cause list to be re-entered): y

 - Run terraform init -upgrade
 
...

Plan: 19 to add, 1 to change, 1 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

...

Apply complete! Resources: 19 added, 1 changed, 1 destroyed.

pgadmin terraform deployment finished. Waiting for pgadmin service (some request failures are expected):
  <urlopen error [Errno 111] Connection refused>. Retrying in 10 seconds...
  HTTP Error 503: Service Temporarily Unavailable. Retrying in 10 seconds...
  HTTP Error 503: Service Temporarily Unavailable. Retrying in 10 seconds...
  HTTP Error 503: Service Temporarily Unavailable. Retrying in 10 seconds...
  HTTP Error 503: Service Temporarily Unavailable. Retrying in 10 seconds...
  HTTP Error 502: Bad Gateway. Retrying in 10 seconds...

pgadmin service is available. URL: https://avritt-dev.civiform.dev:4433
  pgadmin default email: yNCnuDTPPa@default.login
  pgadmin default password: H6gH0bLXvJXO8lItbvs1N5i8YwYYRrG3MS7G8YdC
  postgres database password: default-f!T(]]z&MP*Y^]hw<i8bV{R=Eq{LaoX{zETw$qQa

WARNING: it is strongly recommended to clean up pgadmin resources once they are no-longer needed.
Run 'bin/deploy --tag=latest' to manually clean up pgadmin resources.

Waiting to clean up pgadmin resources.  Press enter to trigger cleanup...

 - Run terraform init -upgrade

...

Plan: 1 to add, 1 to change, 19 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

...

Apply complete! Resources: 1 added, 1 changed, 19 destroyed.
```

### DIFF
--- a/cloud/aws/bin/pgadmin.py
+++ b/cloud/aws/bin/pgadmin.py
@@ -53,8 +53,8 @@ def get_cidr_list() -> str:
 
 def detect_public_ip() -> str:
     try:
-        with urllib.request.urlopen(
-                "https://checkip.amazonaws.com/", timeout=3) as response:
+        with urllib.request.urlopen("https://checkip.amazonaws.com/",
+                                    timeout=3) as response:
             # response contains a newline
             return response.read().decode("ascii").strip()
     except:
--- a/cloud/aws/bin/pgadmin.py
+++ b/cloud/aws/bin/pgadmin.py
@@ -54,7 +54,7 @@ def get_cidr_list() -> str:
 def detect_public_ip() -> str:
     try:
         with urllib.request.urlopen(
-                "https://checkip.amazonaws.com/") as response:
+                "https://checkip.amazonaws.com/", timeout=3) as response:
             # response contains a newline
             return response.read().decode("ascii").strip()
     except:
--- a/cloud/aws/bin/pgadmin.py
+++ b/cloud/aws/bin/pgadmin.py
@@ -1,0 +1,107 @@
+import os
+import sys
+import time
+import re
+import urllib.error
+import urllib.request
+
+from cloud.aws.templates.aws_oidc.bin import resources
+from cloud.aws.templates.aws_oidc.bin.aws_cli import AwsCli
+from cloud.shared.bin.lib import terraform
+
+
+def run(config):
+    os.environ["TF_VAR_pgadmin_cidr_allowlist"] = get_cidr_list()
+    os.environ["TF_VAR_pgadmin"] = "true"
+
+    if not terraform.perform_apply(config):
+        sys.stderr.write("Terraform deployment failed.")
+        raise ValueError("Terraform deployment failed.")
+
+    pgurl = f"{config.get_base_url()}:4433"
+    print(
+        "\npgadmin terraform deployment finished. Waiting for pgadmin service (some request failures are expected):"
+    )
+    wait_for_pgadmin_response(pgurl)
+
+    print(f"\npgadmin service is available. URL: {pgurl}")
+    print_secrets(config)
+
+    input(
+        "\nWARNING: it is strongly recommended to clean up pgadmin resources once they are no-longer needed.\n"
+        "Run 'bin/deploy --tag=latest' to manually clean up pgadmin resources.\n\n"
+        "Waiting to clean up pgadmin resources.  Press enter to trigger cleanup...\n"
+    )
+
+    os.unsetenv("TF_VAR_pgadmin_cidr_allowlist")
+    os.unsetenv("TF_VAR_pgadmin")
+    if not terraform.perform_apply(config):
+        sys.stderr.write("Terraform deployment failed.")
+        raise ValueError("Terraform deployment failed.")
+
+
+def get_cidr_list() -> str:
+    print()
+
+    while True:
+        got = input(
+            "\nREQUIRED: input a comma-separated list of IPv4 CIDR blocks that should have access to the pgadmin service:\n"
+            "To allow a single IP address, use a subnet mask of 32. For example: '192.0.0.1/32'\n"
+            "> ")
+
+        # Validate each block in list.
+        #
+        # Regexp is from cloud/aws/modules/pgadmin/variables.tf.
+        matcher = re.compile(
+            "^([0-9]{1,3}\\.){3}[0-9]{1,3}(\\/([0-9]|[1-2][0-9]|3[0-2]))$")
+        errors = False
+        blocks = [x.strip() for x in got.split(",")]
+        for b in blocks:
+            if matcher.fullmatch(b) == None:
+                print(f"  ERROR: '{b}' is not a valid CIDR block.")
+                errors = True
+        if errors:
+            print("Invalid CIDR blocks found.  Please re-enter list.")
+            continue
+
+        # Terraform expects lists in the `["item1", "item2", ..., "itemN"]` format:
+        # https://developer.hashicorp.com/terraform/language/values/variables#variables-on-the-command-line
+        #
+        # Notably, single quotes are not valid. Strings in lists formatted in f-strings are wrapped in single
+        # quotes so we need to replace them with double quotes.
+        formatted = f"{blocks}".replace("'", '"')
+        print(f"Parsed list: {formatted}")
+        y = input(
+            "\nAccept CIDR list? (anything entered other than 'y' will cause list to be re-entered): "
+        )
+        if y == "y":
+            print()
+            break
+
+    return formatted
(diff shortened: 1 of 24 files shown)


cloud/aws/bin/pgadmin.py
```diff
@@ -0,0 +1,107 @@
...
+        "\npgadmin terraform deployment finished. Waiting for pgadmin service (some request failures are expected):"
+    )
+    wait_for_pgadmin_response(pgurl)
+
+    print(f"\npgadmin service is available. URL: {pgurl}")
+    print_secrets(config)
+
+    input(
+        "\nWARNING: it is strongly recommended to clean up pgadmin resources once they are no-longer needed.\n"
+        "Run 'bin/deploy --tag=latest' to manually clean up pgadmin resources.\n\n"
```
nbeloglazov, line 32: Now I realize that this is potentially problematic. If they run `bin/deploy --tag=latest` then they will deploy latest civiform which should never be used in prod. Instead they should deploy whatever civiform they already have running.

cloud/aws/bin/pgadmin.py
```diff
@@ -0,0 +1,107 @@
...
+        raise ValueError("Terraform deployment failed.")
+
+
+def get_cidr_list() -> str:
+    print()
+
+    while True:
+        got = input(
+            "\nREQUIRED: input a comma-separated list of IPv4 CIDR blocks that should have access to the pgadmin service:\n"
+            "To allow a single IP address, use a subnet mask of 32. For example: '192.0.0.1/32'\n"
```
nbeloglazov, line 49: That need more instructions. Personally I have no idea what to put here. Like I can run `ifconfig` and try to find one of the ips there. But generally that's not correct if I'm sitting behind a NAT and my public IP is different, right? Or provide an option here like "if you ok allowing all ips use 0.0.0.0/..." or whatever is the allow-all address.

nbeloglazov: Maybe there is a way to make a call to some public service that gives you your public ip? Then you can just propose using it.

cloud/shared/bin/pgadmin.py
```diff
@@ -0,0 +1,19 @@
+import subprocess
+import importlib
+import os
+from typing import List
+
+from cloud.shared.bin.lib.config_loader import ConfigLoader
+
+
+def run(config: ConfigLoader, params: List[str]):
```
nbeloglazov, line 9: To launch pgadmin the human admin must provide `--tag=...` option, right? Which means they need to use exact version of the app that is currently deployed. Otherwise they will accidentally re-deploy civiform which is pretty undesirable. Can we detect it somehow and prevent? Initially I thought we should just forbid using "--tag=latest" with pgadmin but then remembered that we want it to work on staging as well.

Maybe we can extract previous version from terraform state? Alternatively we could try hacks like reading currently deployed version from civiform webpage html.

nbeloglazov (changes requested):

cloud/aws/bin/pgadmin.py
avrittrohwer: Done.

cloud/aws/bin/pgadmin.py
avrittrohwer: Updated the output to use whatever --tag the user passed to /bin/run

cloud/aws/bin/pgadmin.py
```diff
@@ -0,0 +1,191 @@
...
+from cloud.aws.templates.aws_oidc.bin import resources
+from cloud.aws.templates.aws_oidc.bin.aws_cli import AwsCli
+
+
+def run(config: ConfigLoader):
+    os.environ["TF_VAR_pgadmin_cidr_allowlist"] = get_cidr_list()
+    os.environ["TF_VAR_pgadmin"] = "true"
+    run_terraform(config)
+
+    pgurl = f"{config.get_base_url()}:4433"
```
bion, line 22: nit: `pg_url` is easier to read

cloud/aws/bin/pgadmin.py
```diff
@@ -0,0 +1,191 @@
...
+        f"Run 'bin/deploy --tag={os.environ['TF_VAR_image_tag']}' to manually clean up pgadmin resources.\n\n"
+        "Waiting to clean up pgadmin resources.  Press enter to trigger cleanup...\n"
+    )
+
+    os.unsetenv("TF_VAR_pgadmin_cidr_allowlist")
+    os.unsetenv("TF_VAR_pgadmin")
+    run_terraform(config)
+
+
+def get_cidr_list() -> str:
```
bion, line 42: suggestion (blocking): it's not immediately apparent which of these free-floating functions are intended to be public and which are not. Please prepend a single underscore to the function names that aren't intended to be called outside of this file.

Additionally, code organization here would benefit from another class that encapsulates most of these free-floating functions.

cloud/aws/bin/pgadmin.py
```diff
@@ -0,0 +1,191 @@
...
+    while True:
+        prompt = sm.next(user_input)
+        if prompt == "":
+            return sm.cidrs()
+
+        user_input = input(prompt)
+
+
+def detect_public_ip() -> str:
+    """Code from https://api.ident.me/"""
```
bion, line 55: question (blocking): if this service becomes unavailable, what happens?

bion (changes requested):

cloud/aws/bin/pgadmin.py
avrittrohwer: I updated it to call checkip.amazonaws.com and added a timeout.  So if the call fails, an empty string will be returned and the user will be prompted to input CIDR blocks

cloud/aws/bin/pgadmin.py
avrittrohwer: Non of these functions except for run() are expected to be used outside of this file.  Added leading _

cloud/shared/bin/pgadmin.py
avrittrohwer: What do you think about the approach in https://github.com/civiform/civiform-deploy/pull/30?  It avoids needing to dynamically detect anything

bion (approved):

cloud/aws/bin/pgadmin_test.py
```diff
@@ -0,0 +1,113 @@
...
+
+ To run the tests: PYTHONPATH="${PYTHONPATH}:${pwd}" python3 cloud/aws/bin/pgadmin_test.py
+"""
+
+
+def ip(ip):
+    return lambda: ip
+
+
+class TestCIDRInputStateMachine(unittest.TestCase):
```
nbeloglazov, line 15: Do we have github action that runs unit tests for this repo?

cloud/shared/bin/pgadmin.py
nbeloglazov: Looks good!

nbeloglazov (approved):

cloud/aws/bin/pgadmin_test.py
avrittrohwer: Yes, we have the `PyTest / pytest (pull_request)` check which shows running the test file in its output:

```
cloud/aws/bin/pgadmin_test.py ........                                   [ 25%]
```

PR #99, opened 2022-12-20
Update print function to write to stderr

avrittrohwer:
Fixes https://github.com/civiform/civiform/issues/3967.

### DIFF
--- a/cloud/shared/bin/lib/print.py
+++ b/cloud/shared/bin/lib/print.py
@@ -1,8 +1,8 @@
 import builtins
 import sys
 
+
 def print(*args, **kwargs):
     kwargs["file"] = sys.stderr
     kwargs["flush"] = True
     builtins.print(*args, **kwargs)
-
--- a/cloud/shared/bin/lib/print.py
+++ b/cloud/shared/bin/lib/print.py
@@ -1,0 +1,8 @@
+import builtins
+import sys
+
+def print(*args, **kwargs):
+    kwargs["file"] = sys.stderr
+    kwargs["flush"] = True
+    builtins.print(*args, **kwargs)
+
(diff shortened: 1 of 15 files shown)


bion (changes requested):
Please resolving failing CI checks

avrittrohwer:
Fixed checks

cloud/shared/bin/lib/print.py
```diff
@@ -0,0 +1,8 @@
+import builtins
+import sys
+
+
+def print(*args, **kwargs):
```
bion, line 5: suggestion (blocking): rename to `print_error` so it's clear this goes to stderr, we should be able to print info output to stdout

bion (changes requested):

avrittrohwer:
The reason I shadowed print is that this diff for this PR is way smaller.
I'm not sure what the deploy tool would use stdout for, I can't imagine
piping the deploy tool to anything other than logging.  In the future if we
need to print to stdout, we can change the shadowed print function to not
override the file named parameter and callers could pass file=sys.stdout.
Thoughts?

cloud/shared/bin/lib/print.py
avrittrohwer: The reason I shadowed print is that this diff for this PR is way smaller.
I'm not sure what the deploy tool would use stdout for, I can't imagine
piping the deploy tool to anything other than logging.  In the future if we
need to print to stdout, we can change the shadowed print function to not
override the file named parameter and callers could pass file=sys.stdout.
Thoughts?

bion: Sorry for the slow response!

Printing error messages to stdout is fine I think. Some people may have their terminals configured to print stderr in a different color or something but it's not a big deal.

We are however already printing to stderr in a few other places, so it'd be good to get consistent.

bion (approved):

PR #128, opened 2023-02-28
Add reporting UI feature flag

bion:
Feature flag for admin reporting UI.

### DIFF
--- a/cloud/azure/modules/app/locals.tf
+++ b/cloud/azure/modules/app/locals.tf
@@ -79,6 +79,7 @@ locals {
 
     CIVIFORM_API_SECRET_SALT = data.azurerm_key_vault_secret.api_secret_salt_key.value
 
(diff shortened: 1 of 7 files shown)


cloud/azure/modules/app/locals.tf
```diff
@@ -79,6 +79,7 @@ locals {
 
     CIVIFORM_API_SECRET_SALT = data.azurerm_key_vault_secret.api_secret_salt_key.value
 
+    CIVIFORM_ADMIN_REPORTING_UI_ENABLED          = var.feature_flag_reporting_enabled
```
avrittrohwer, line 82: Where is this used?

bion: It's setting the env var for the server

avrittrohwer: I see, I missed it is in the `app_settings` object

avrittrohwer (approved):

PR #130, opened 2023-03-08
Improve test coverage and documentation for variable loading and validation

jessbumblebee:
- Adds test for the variable_definition_loader
- Removes some unused code that refers to a method that no longer exists
- Adds a comment that links to an existing bug for some suggested changes to validation for the purpose of saving future developers time to figure out that variable definition validation is currently executed by github actions only.
- Removes additional bit of code that does nothing

### DIFF
--- a/cloud/shared/bin/lib/variable_definition_loader.py
+++ b/cloud/shared/bin/lib/variable_definition_loader.py
@@ -12,8 +12,8 @@ class VariableDefinitionLoader:
 
     def load_definition_file(self, definition_file_path: str):
         with open(definition_file_path, "r") as file:
-            # json.loads() returns a dictionary, keeping the last 
-            # occurance and thus removes duplicates. 
+            # json.loads() returns a dictionary, keeping the last
+            # occurance and thus removes duplicates.
             # To detect them we would need to write our own parser.
             definitions = json.loads(file.read())
 
--- a/cloud/shared/bin/lib/variable_definition_loader.py
+++ b/cloud/shared/bin/lib/variable_definition_loader.py
@@ -12,7 +12,8 @@ class VariableDefinitionLoader:
 
     def load_definition_file(self, definition_file_path: str):
         with open(definition_file_path, "r") as file:
-            # json.loads() returns a dictionary and thus removes duplicates. 
+            # json.loads() returns a dictionary, keeping the last 
+            # occurance and thus removes duplicates. 
             # To detect them we would need to write our own parser.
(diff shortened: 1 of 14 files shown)


cloud/shared/bin/lib/variable_definition_loader_test.py
```diff
@@ -0,0 +1,50 @@
...
+            },
+            "BOB": {
+                "required": False,
+                "secret": False,
+                "type": "string"
+            },
+        }
+
+    def test_load_valid_variable_definitions(self):   
+        self.__json_file_path_name = "./valid_vars_testfile.json"
```
avrittrohwer, line 36: You can use tempfile.NamedTemporaryFile() to create these files: https://docs.python.org/3/library/tempfile.html

```
with tempfile.NamedTemporaryFile(mode="w") as f:
    self._write_json_file(self.__valid_variable_defs, f)
    varDefLoader.load_definition_file(f.name)
...
```

cloud/shared/bin/lib/variable_definition_loader.py
```diff
@@ -2,26 +2,21 @@
...
+    """ 
+    Loads all variables from a valid variable definition json file.
+    """
 
     def __init__(self, variable_definitions={}):
         self.variable_definitions: dict = variable_definitions
 
     def load_definition_file(self, definition_file_path: str):
         with open(definition_file_path, "r") as file:
+            # json.loads() returns a dictionary and thus removes duplicates.
```
avaleske, line 15: Does it keep the last one it sees or the first one?

invalid_vars_testfile.json
```diff
@@ -0,0 +1 @@
+{"BING": {"required": true, "secret": false, "type": "string"}, "BONG": {"required": true, "secret": false, "type": "string"}, "BONG": {"required": false, "secret": false, "type": "string"}}
```
avaleske, line 1: nit: newlines here and in the other file

avaleske (approved):

cloud/shared/bin/lib/variable_definition_loader.py
jessbumblebee: It keeps the last one (comment updated)

invalid_vars_testfile.json
jessbumblebee: I removed the file because it was actually an accidental leftover. I created it dynamically and removed the line to delete it to see the content for manual testing.

PR #133, opened 2023-03-16
Set CIVIFORM_VERSION instead of CIVIFORM_IMAGE_TAG

avrittrohwer:
CIVIFORM_VERSION was added in https://github.com/civiform/civiform/pull/4405.  CIVIFORM_IMAGE_TAG is set in the prod.Dockerfile: https://github.com/civiform/civiform/blob/main/prod.Dockerfile#L46.

This change makes it so if CIVIFORM_VERSION in the civiform_config.sh file is set to "latest", the image tag (SNAPSHOT-3af89974-1678907834 for example) is show on the login page when SHOW_CIVIFORM_IMAGE_TAG_ON_LANDING_PAGE is enabled instead of "latest".  If CIVIFORM_VERSION is a release version like v1.18.0, only the release version is shown.

Fixes https://github.com/civiform/civiform/issues/4270.

Tested: Set CIVIFORM_VERSION to "latest" and deployed in my dev account, snapshot version is shown: 
image

### DIFF
--- a/cloud/aws/templates/aws_oidc/app.tf
+++ b/cloud/aws/templates/aws_oidc/app.tf
@@ -84,7 +84,7 @@ module "civiform_server_container_def" {
     STORAGE_SERVICE_NAME = "s3"
     AWS_S3_BUCKET_NAME   = aws_s3_bucket.civiform_files_s3.id
 
-    CIVIFORM_IMAGE_TAG                      = var.image_tag
+    CIVIFORM_VERSION                        = var.image_tag
     SHOW_CIVIFORM_IMAGE_TAG_ON_LANDING_PAGE = var.show_civiform_image_tag_on_landing_page
     CIVIFORM_TIME_ZONE_ID                   = var.civiform_time_zone_id
     WHITELABEL_CIVIC_ENTITY_SHORT_NAME      = var.civic_entity_short_name


cloud/aws/templates/aws_oidc/app.tf
```diff
@@ -84,7 +84,7 @@ module "civiform_server_container_def" {
     STORAGE_SERVICE_NAME = "s3"
     AWS_S3_BUCKET_NAME   = aws_s3_bucket.civiform_files_s3.id
 
-    CIVIFORM_IMAGE_TAG                      = var.image_tag
+    CIVIFORM_VERSION                        = var.image_tag
```
rockycodes, line 87: Don't we still need `CIVIFORM_IMAGE_TAG` for use [here](https://github.com/civiform/civiform/blob/b556a3343e0c62a6f928f8750d8c1dd50a80919b/server/conf/application.conf#L485)? Or am I misunderstanding how it's hooked up?

avrittrohwer: The [bin/build-prod](https://github.com/civiform/civiform/blob/main/bin/build-prod) script builds the civiform/civiform docker image we push to dockerhub.  It uses [prod.Dockerfile](https://github.com/civiform/civiform/blob/main/prod.Dockerfile#L46) to build the image.

In prod.Dockerfile, we have `ENV CIVIFORM_IMAGE_TAG=$image_tag` which sets the CIVIFORM_IMAGE_TAG environment variable inside of the docker container.  I previously added the `CIVIFORM_IMAGE_TAG = var.image_tag` line so that we can set the version to a nice version string like 'v1.18.0'.  This change adds a new CIVIFORM_VERSION to pass the version string instead of overwriting the image tag variable

rockycodes (approved):

PR #134, opened 2023-03-17
Reading configs directly from .sh rather than env variables

jessbumblebee:
Reading configs directly from .sh rather than env variables

More information here: https://docs.google.com/document/d/1M27DF1IJypaeyLQrTZAy-y2DEszUV-EgGkhuNgOMzRE/edit#heading=h.bknox3ib8wpb

Avritt's change touches some of the same files (config loader). As we are both on and off this week, I suggest that whoever is ready to submit first submits and the other does the merge, so that we both stay unblocked. I had a look and the merge looks similarly feasible no matter which change goes in first.


cloud/shared/bin/lib/config_parser.py
```diff
@@ -0,0 +1,53 @@
...
+                        )
+                    # extract the key value pairs without the "export"
+                    key_and_value_with_comments = line[len(export_string
+                                                          ):].strip().split(
+                                                              "=", 1)
+
+                    var_name = key_and_value_with_comments[0].strip()
+
+                    # strip out in line comments and remove quotes and blanks
+                    value = key_and_value_with_comments[1].split("#", 1)[0]
```
bion, line 40: **issue:** this is a little tough. we could have quoted `#` in things like URLs that contain fragments. recommend splitting on `\w+#` since we should be able to expect that inline comments have a space preceding them?

bion (approved):

PR #143, opened 2023-04-12
Allow deployer to set DATABASE_APPLY_DESTRUCTIVE_CHANGES

bion:
When rolling back to a previous release, the server needs to be synchronized with the database schema it is expecting. It does this by applying the `downs` in the [evolutions](https://www.playframework.com/documentation/2.8.x/Evolutions) for the DB versions beyond the one it is expecting. This is a destructive action, since it involves removing things added in the `ups` such as tables, indices, etc.

By default in production mode, Play does not allow destructive `downs`. However it may be necessary to destroy data (for example, a recently created column when rolling back a bad release) when mitigating a production incident.

This PR adds `DATABASE_APPLY_DESTRUCTIVE_CHANGES` as a variable the deployer can set in their civiform_config.sh file. It corresponds with the existing variable in the [server conf](https://github.com/civiform/civiform/blob/main/server/conf/application.conf#L407-L409).

addresses https://github.com/civiform/civiform/issues/2236


rockycodes (commented):
This looks good to me, but I'm going to let Jessica approve since she's more familiar with passing variables through. 

Just curious though, is this connected to an issue or what inspired this work? Is it weird that we have it referenced in the server conf but weren't passing it through here already or how was it being set/referenced previously? 

Also just confirming you will do the follow-up work to add this to the [sample configuration file](https://github.com/civiform/civiform-deploy/blob/main/civiform_config.example.sh).

bion:
@rockycodes 

> Just curious though, is this connected to an issue or what inspired this work? 

https://github.com/civiform/civiform/issues/2236

> Is it weird that we have it referenced in the server conf but weren't passing it through here already or how was it being set/referenced previously?

It wasn't settable previously using the deployment system. The `DATABASE_APPLY_DESTRUCTIVE_CHANGES` env var in the server config predates the deployment system.

bion:
@rockycodes 

> Also just confirming you will do the follow-up work to add this to the [sample configuration file](https://github.com/civiform/civiform-deploy/blob/main/civiform_config.example.sh).

yep!

bion:
No longer needed thanks to auto-generating TF vars

PR #132, opened 2023-03-15
Support setting CiviForm server environment variables directly

avrittrohwer:
Supports setting CiviForm server environment variables defined in civiform/civiform server/conf/env-var-docs.json without defining them in the variable_definitions.json files in this repo.

This functionality is enabled by the env_var_docs.parser package being available to import.  Support for making that package available is implemented in https://github.com/civiform/civiform-deploy/pull/34/files.  If the package is not importable, this change is a no-op.

TODO:

- [ ] Read env-var-docs.json file from github based on CIVIFORM_VERSION.

Tested:

In my config file, I set:

```
export CIVIFORM_SUPPORTED_LANGUAGES="en-US, am, zh-TW"
```

This resulted in the generated .tfvars file having:

```
civiform_server_environment_variables = {
  "LOGIN_RADIUS_METADATA_URI"=""
  "STAGING_HOSTNAME"="avritt-dev.civiform.dev"
  "BASE_URL"="https://avritt-dev.civiform.dev"
  "APPLICANT_OIDC_LOCALE_ATTRIBUTE"=""
  "APPLICANT_OIDC_MIDDLE_NAME_ATTRIBUTE"=""
  "APPLICANT_OIDC_FIRST_NAME_ATTRIBUTE"="name"
  "AWS_REGION"="us-east-1"
  "APPLICANT_OIDC_LAST_NAME_ATTRIBUTE"=""
  "LOGIN_RADIUS_API_KEY"=""
  "ADFS_ADDITIONAL_SCOPES"="allatclaims"
  "CIVIFORM_APPLICANT_IDP"="generic-oidc"
  "CIVIFORM_SUPPORTED_LANGUAGES.0"="en-US"
  "CIVIFORM_SUPPORTED_LANGUAGES.1"="am"
  "CIVIFORM_SUPPORTED_LANGUAGES.2"="zh-TW"
  "APPLICANT_OIDC_RESPONSE_MODE"="form_post"
  "CIVIFORM_TIME_ZONE_ID"="America/Los_Angeles"
  "APPLICANT_OIDC_ADDITIONAL_SCOPES"=""
  "APPLICANT_OIDC_DISCOVERY_URI"="https://civiform-staging.us.auth0.com/.well-known/openid-configuration"
  "APPLICANT_OIDC_PROVIDER_NAME"="OidcClient"
  "APPLICANT_OIDC_EMAIL_ATTRIBUTE"="email"
[...]

PR #147, opened 2023-04-24
Add the plumbing for enabling variable auto generation via env var docs

jessbumblebee:
This PR adds a new script "run" which runs before run.py.
The purpose is to download and install the env-var-docs python package.
We create a python virtual environment(venv), install the dynamically downloaded env-var-docs package in it and run our python deploy code inside the venv so it has access to env-var-docs.

#132 contains the feature this would enable, whose longer term goal is to automatically generate the variables that allow configuring the server via the deploy system.
Note that this PR does not yet enable the feature. It merely provides the script that can do so. 
When we are happy to enable the feature, we should replace all calls to run.py with run, which then calls run.py in turn.

Some reasoning about the location of this: 
We need to enable the python packages across all deployment implement system (dockerized deployment, checkout deployment, AWS. Azure). The run.py script is the first step that is shared between all those options. There are some other benefits to placing it here rather than somewhere in the deploy repository (e.g. you can use run during development to not have to code across multiple branches) 
We can not enable the package in run.py directly, because the activation of the python virtual environment needs to happen in the parent process which the python process is started from before the code is started.


cloud/shared/bin/python_env_setup
```diff
@@ -0,0 +1,54 @@
...
+  local requirements_file_path=$1
+
+  # Check if there are any requirements to install
+  if [[ ! -f "$requirements_file_path" ]]; then
+    echo "Requirements file $requirements_file_path does not exist, python setup skipped"
+    return
+  fi
+
+  # Create a virtual environment if it doesn't exist yet
+  if [[ ! -d .venv ]]; then
```
avrittrohwer, line 25: The scripts in civiform-deploy remove the checkout directory on every run: https://github.com/civiform/civiform-deploy/blob/main/bin/lib/checkout.sh#L80.  This was due to the checkout directory getting larger on each run: https://github.com/civiform/civiform-deploy/pull/31.

If we put the venv directory in the checkout directory, with the current civiform-deploy scripts, we will re-download the env-var-docs parser package on every deployment tool run.  This adds 5-10 seconds of latency to every run.  What do you think about modifying the civiform-deploy script to not remove the .venv directory?

cloud/shared/bin/run
```diff
@@ -0,0 +1,30 @@
+#! /usr/bin/env bash
+# Script for running the commands supported by the deploy system by calling run.py.
+# See comments below for why we need this scripts rather than running run.py directly.
+# TODO(#4612) Enable the env-var-docs variable generation feature by using this script
+# instead of run.py across the civiform-deploy repository
+
+# In the server code repository in git(civiform), we define a python package called env-var-docs,
```
rockycodes, line 7: suggestion (non-blocking): link to package in comment https://github.com/civiform/civiform/tree/main/env-var-docs

cloud/shared/bin/run
```diff
@@ -0,0 +1,30 @@
...
+# whose purpose is to automate various aspects of adding a variable to application.conf.
+# Variables in applicaiton.conf make the server configurable.
```
rockycodes, line 9:
```suggestion
# Variables in application.conf make the server configurable.
```

cloud/shared/bin/run
```diff
@@ -0,0 +1,30 @@
+#! /usr/bin/env bash
+# Script for running the commands supported by the deploy system by calling run.py.
+# See comments below for why we need this scripts rather than running run.py directly.
```
rockycodes, line 3:
```suggestion
# See comments below for why we need this script rather than running run.py directly.
```

rockycodes (commented):
Looks good to me! I don't feel like I understand it quite well enough to approve, so I leave that to Avritt, but I left a couple of comments for readability.

cloud/shared/bin/python_env_setup
jessbumblebee: Good catch, I'll look into it in a separate PR as it needs to be done in the other repository.
We could either exclude the .venv from the removal or place the .venv elsewhere, I'll have a think about it.

cloud/shared/bin/run
jessbumblebee: Done

cloud/shared/bin/run
jessbumblebee: Done.

cloud/shared/bin/run
jessbumblebee: Done.

cloud/shared/bin/python_env_setup
jessbumblebee: Ps: as the PR does not yet enable the feature, the latency won't have an impact.

avrittrohwer (approved):

PR #148, opened 2023-04-26
Add ability to restore DB to snapshot

bion:
Adds `POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER` config var which can be used to destroy the replace the database with a new one restored from the specified snapshot.

The RDS instance has deletion protection on my default, requiring the deployer to access the AWS console and disable it before this can be used.

### DIFF
--- a/cloud/aws/templates/aws_oidc/main.tf
+++ b/cloud/aws/templates/aws_oidc/main.tf
@@ -29,6 +29,9 @@ resource "aws_db_instance" "civiform" {
     Name = "${var.app_prefix} Civiform Database"
     Type = "Civiform Database"
   }
+
+  # If not null, destroys the current database, replacing it with a new one restored from the provided snapshot
+  snapshot_identifier             = var.postgres_restore_snapshot_identifier
   deletion_protection             = local.deletion_protection
   instance_class                  = var.postgres_instance_class
   allocated_storage               = var.postgres_storage_gb
--- a/cloud/aws/templates/aws_oidc/variable_definitions.json
+++ b/cloud/aws/templates/aws_oidc/variable_definitions.json
@@ -184,5 +184,11 @@
     "secret": false,
     "tfvar": true,
     "type": "integer"
+  },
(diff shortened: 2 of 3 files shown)


cloud/aws/templates/aws_oidc/main.tf
```diff
@@ -29,6 +29,9 @@ resource "aws_db_instance" "civiform" {
     Name = "${var.app_prefix} Civiform Database"
     Type = "Civiform Database"
   }
+
+  # If not null, destroys the current database, replacing it with a new one restored from the provided snapshot
```
gwendolyngoetz, line 33: Do we need this comment here given it's the same as the variable description? I worry they could get out of sync.

gwendolyngoetz: I find my eyes just slide past it if looking at the whole file with it being a single line comment. You indicated that there's a manual step needing to be done in the AWS console to allow it to run. That might be enough of a safeguard.

Oh and now that I think about it. It's probably unlikely the civic entity doing the deploying will look at the comment in that file.

gwendolyngoetz (approved):
Seems good to me. I'll leave it up to you @bion if you want to remove that comment or not.

dkatzz (approved):

rockycodes (approved):
LGTM, but I feel like I'm missing some context. Can you explain more about who would do this under what circumstances and how? Is the intention that `POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER` will be included in the `civiform_config` and and IT person could set it to a specific snapshot and then re-deploy?

bion:
@rockycodes

> Is the intention that `POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER` will be included in the `civiform_config` and and IT person could set it to a specific snapshot and then re-deploy?

Yes

> Can you explain more about who would do this under what circumstances and how?

A deployment maintainer would use this to reset their database to a previous snapshot in the event of unintended data loss or widespread data corruption.

I'll be adding a playbook for how/when to use this in the docs repo.

PR #151, opened 2023-04-26
Set ownership controls on access log bucket

bion:
Followed instructions in https://github.com/hashicorp/terraform-provider-aws/issues/28353 to fix https://github.com/civiform/civiform/issues/4697

This PR also disables storage of load balancer logs on S3. The module we use for creating the fargate service has been setting up an S3 bucket for load balancer logs behind the scenes. We've never had need to view those logs, and don't have cloudwatch set up so that we could look at them without downloading and un-gzipping them directly from S3, which we've never done.

If we merge this PR with LB logs disabled, existing deployments will need to empty their LB log bucket before they'll be able to deploy again, otherwise terraform will fail with an error stating that it can't delete a non-empty bucket.

Alternatively, we could wait for this PR to merge and then update the module version [here](https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/aws/templates/aws_oidc/app.tf#L341) once the change has propagated through the intermediate dependencies.

Ran bin/setup and verified the bug ACL error is no longer occurring.

### DIFF
--- a/cloud/aws/templates/aws_oidc/app.tf
+++ b/cloud/aws/templates/aws_oidc/app.tf
@@ -350,6 +350,7 @@ module "ecs_fargate_service" {
   ecs_cluster_arn         = module.ecs_cluster.aws_ecs_cluster_cluster_arn
   private_subnets         = module.vpc.private_subnets
   public_subnets          = module.vpc.public_subnets
+  enable_s3_logs          = false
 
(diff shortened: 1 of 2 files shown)


rockycodes (approved):
🎉

gwendolyngoetz (approved):

PR #164, opened 2023-05-04
Add ESRI service to AWS staging

rockycodes:
This PR pipes through the variables we need to enable Esri address correction in staging for AWS. Service area validation vars are not included here because they need to be sent through as arrays/lists which we don't have a pattern for yet (Issue to track that work is here: https://github.com/civiform/civiform/issues/4790). Eventually, we will need to include [all Esri vars](https://github.com/civiform/civiform/blob/fd0aaa002e2ee01d378ca90f236c316641ed0101/server/conf/application.conf#L722) in the deploy system.

Issue: https://github.com/civiform/civiform/issues/4621
Corresponding PR to add these to our AWS staging config file: https://github.com/civiform/civiform-staging-deploy/pull/47

### DIFF
--- a/cloud/aws/templates/aws_oidc/app.tf
+++ b/cloud/aws/templates/aws_oidc/app.tf
@@ -132,8 +132,8 @@ module "civiform_server_container_def" {
     COMMON_INTAKE_MORE_RESOURCES_LINK_TEXT = var.common_intake_more_resources_link_text
     COMMON_INTAKE_MORE_RESOURCES_LINK_HREF = var.common_intake_more_resources_link_href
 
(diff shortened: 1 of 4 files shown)


gwendolyngoetz (approved):

gwendolyngoetz (approved):

rockycodes:
@gwendolyngoetz that's good to know. I was going to check in with you about that. I time boxed figuring out how to pipe arrays through the deploy system this morning and wasn't able to do it. I'm going to merge this to unblock some perf measurement stuff @dkatzz is working on and will make a new issue for the service area vars. I'll bring it up at Monday's meeting as well.

PR #183, opened 2023-05-23
download correct version of parser package for civiform version

rockycodes:
Now that [we have logic](https://github.com/civiform/cloud-deploy-infra/pull/154) to pull in the correct version of `env-var-docs.json` for the version of CiviForm being deployed, we need to make sure we're pulling in the correct version of the parser package as well.

This change adds logic to the `run` script to fetch the correct version of the parser package via the Github api based on the version of CiviForm being deployed.

I tested it locally by deploying with different versions (v1.24.0 and "latest") and confirming that different versions of the parser packaged were downloaded.

### DIFF
--- a/cloud/shared/bin/run
+++ b/cloud/shared/bin/run
@@ -32,11 +32,12 @@ if [ $tag == "latest" ]; then
     | python3 -c "import sys, json; print(json.load(sys.stdin)['tag_name'])")
 fi
 
-# Get the correct version of the env-var-docs/parser-package and write it to the dependencies file for installation
+# Get the commit sha for the commit at the tip of the CiviForm version being deployed
 tag_url=$(curl -s "https://api.github.com/repos/civiform/civiform/git/refs/tags/${tag}" \
   | python3 -c "import sys, json; print(json.load(sys.stdin)['object']['url'])")
 commit_sha=$(curl -s "${tag_url}" \
   | python3 -c "import sys, json; print(json.load(sys.stdin)['object']['sha'])")
+# Write the correct version of the env-var-docs/parser-package to the dependencies file to be downloaded
 echo "env-var-docs @ git+https://github.com/civiform/civiform.git@${commit_sha}#subdirectory=env-var-docs/parser-package" >>"cloud/shared/bin/env-var-docs-python-dependencies.txt"
 
 initialize_python_env cloud/shared/bin/env-var-docs-python-dependencies.txt
--- a/cloud/shared/bin/run
+++ b/cloud/shared/bin/run
@@ -28,16 +28,16 @@ done
 
 # If tag is "latest" get the latest release number (eg. v1.26.0)
 if [ $tag == "latest" ]; then
-  tag=`curl -s "https://api.github.com/repos/civiform/civiform/releases/latest" | \
-    python3 -c "import sys, json; print(json.load(sys.stdin)['tag_name'])"`
+  tag=$(curl -s "https://api.github.com/repos/civiform/civiform/releases/latest" \
+    | python3 -c "import sys, json; print(json.load(sys.stdin)['tag_name'])")
 fi
 
 # Get the correct version of the env-var-docs/parser-package and write it to the dependencies file for installation
-tag_url=`curl -s "https://api.github.com/repos/civiform/civiform/git/refs/tags/${tag}" | \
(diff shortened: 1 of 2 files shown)


cloud/shared/bin/run
```diff
@@ -13,7 +13,33 @@
...
+    s) sourceconfig=${OPTARG} ;;
+    # The command that the run.py script should execute
+    c) command=${OPTARG} ;;
+    # The tag of the image that should be used for this deployment (e.g. "latest")
+    t) tag=${OPTARG} ;;
+  esac
+done
+
+# If tag is "latest" get the latest release number (eg. v1.26.0)
+if [ $tag == "latest" ]; then
```
bion, line 30: **nitpick:** https://google.github.io/styleguide/shellguide.html#s6.3-tests
```suggestion
if [[ ${tag} == "latest" ]]; then
```

cloud/shared/bin/run
```diff
@@ -13,7 +13,33 @@
...
+  tag=$(curl -s "https://api.github.com/repos/civiform/civiform/releases/latest" \
+    | python3 -c "import sys, json; print(json.load(sys.stdin)['tag_name'])")
+fi
+
+# Get the commit sha for the commit at the tip of the CiviForm version being deployed
+tag_url=$(curl -s "https://api.github.com/repos/civiform/civiform/git/refs/tags/${tag}" \
+  | python3 -c "import sys, json; print(json.load(sys.stdin)['object']['url'])")
+commit_sha=$(curl -s "${tag_url}" \
+  | python3 -c "import sys, json; print(json.load(sys.stdin)['object']['sha'])")
```
bion, line 39: **suggestion (blocking):** please factor out a function for this repeated pattern of piping the output of `curl` through python's JSON lib. E.g.

```shell
#######################################
# Get the value in a JSON document at the given URL.
# Arguments:
#   1: URL of the JSON document
#   2: Path of the value in bracket notation e.g. "['object']['sha']"
# Returns:
#   Value found at the specified path in the JSON document
#######################################
function fetch_json_val() {
  curl -s "${1}" \
    python3 -c "import sys, json; print(json.load(sys.stdin)${2})"
}
```

cloud/shared/bin/run
```diff
@@ -13,7 +13,33 @@
 # activate the venv. The activation has to happen in the parent process before the python code is run.
 # This is why we need a bash script that preceeds the python code
 source cloud/shared/bin/python_env_setup
-# TODO(#4612)Download the newest version of the package rather than a fixed version.
+
+# Get the arguments that we want to pass to run.py
+while getopts s:c:t: flag; do
+  case "${flag}" in
+    # The civiform_config file that contains the values to configure the deployment
+    s) sourceconfig=${OPTARG} ;;
```
bion, line 21: **nitpick:** 
```suggestion
    s) source_config=${OPTARG} ;;
```

bion (changes requested):

cloud/shared/bin/run
rockycodes: done. thanks for the suggested code!

cloud/shared/bin/run
```diff
@@ -13,7 +13,47 @@
...
+    c) command=${OPTARG} ;;
+    # The tag of the image that should be used for this deployment (e.g. "latest")
+    t) tag=${OPTARG} ;;
+  esac
+done
+
+# If tag is "latest" get the latest release number (eg. v1.26.0)
+if [[ ${tag} == "latest" ]]; then
+  tag=$(curl -s "https://api.github.com/repos/civiform/civiform/releases/latest" \
+    | python3 -c "import sys, json; print(json.load(sys.stdin)['tag_name'])")
```
bion, line 32: **nitpick:** `fetch_json_val` here as well

cloud/shared/bin/run
bion, line 1: **nitpick:** throughout this file please reduce line length to 80 chars max: https://google.github.io/styleguide/shellguide.html#line-length-and-long-strings

bion (approved):

cloud/shared/bin/run
rockycodes: oh shoot i just merged before i saw these requested changes. i'll make these edits in the next pr.

rockycodes: we have a shell formatter that runs. i'll see if there's an option to limit line length.

PR #192, opened 2023-06-12
Add ECS task CPU and memory variables

dkatzz:
Allow configuration of ECS task CPU and memory

https://github.com/civiform/civiform/issues/5058

### DIFF
--- a/cloud/shared/variable_definitions.json
+++ b/cloud/shared/variable_definitions.json
@@ -260,17 +260,5 @@
     "secret": false,
     "tfvar": true,
     "type": "bool"
-  },
-  "ECS_TASK_CPU": {
-    "required": false,
-    "secret": false,
-    "tfvar": true,
-    "type": "integer"
-  },
-  "ECS_TASK_MEMORY": {
-    "required": false,
-    "secret": false,
-    "tfvar": true,
-    "type": "integer"
   }
 }
--- a/cloud/shared/variable_definitions.json
+++ b/cloud/shared/variable_definitions.json
@@ -260,5 +260,17 @@
     "secret": false,
     "tfvar": true,
     "type": "bool"
+  },
+  "ECS_TASK_CPU": {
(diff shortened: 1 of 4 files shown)


bion (changes requested):
Probably also need to add here https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/shared/variable_definitions.json ?

dkatzz:
> Probably also need to add here https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/shared/variable_definitions.json ?

Added, but it's kind of unclear to me when we put things in that file and when we don't? I noticed other AWS configuration values like `postgres_backup_retention_days` and `postgres_instance_class` aren't included, but some are. Should every variable be in there or do we have some sort of rule for when things should be included and when they shouldn't?

bion:
@dkatzz 

> Should every variable be in there 

yes

bion (approved):

cloud/shared/variable_definitions.json
```diff
@@ -260,5 +260,17 @@
...
+    "required": false,
+    "secret": false,
+    "tfvar": true,
+    "type": "integer"
+  },
+  "ECS_TASK_MEMORY": {
+    "required": false,
+    "secret": false,
+    "tfvar": true,
+    "type": "integer"
```
bion, line 274: **suggestion:** ah woops, sorry. I linked to the wrong file. This should be in the AWS-specific variable definitions: https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/aws/templates/aws_oidc/variable_definitions.json

bion (changes requested):

cloud/shared/variable_definitions.json
dkatzz: Ahh makes sense, thank you!!

bion (approved):

PR #195, opened 2023-06-14
Remove use of ecs-fargate-service module

nb1701:
This module, and particularly its dependency on ecs-alb, has caused us many problems. This takes the resources that were defined by these modules and uses them in our own version of a module, streamlined to just include the resources we need. Additionally, the loadbalancer target group names were reduced in size, since we were often hitting the 32 character limit with the naming scheme used in ecs-alb and using the app prefix with "-civiform" appended to it.  Most resources have a 63+ character limit, but target group names have 32.

Fixes https://github.com/civiform/civiform/issues/4700, https://github.com/civiform/civiform/issues/4826


bion (approved):
Really glad to see this! I almost started pointing out opportunities for simplification but agree that doing that in a separate PR is preferable

bion:
@nb1701 

> I think I might take a different path with this, and pull in the module code and keep it as a module within this repo, so hopefully it appears to TF that it's the same resource. I don't love forking the module like that, but that's basically what we're doing here, and we could also optimize out the stuff we aren't using to streamline it. If that doesn't work, I'll have to probably rename all the new resources :( Which maybe isn't so bad, we could remove the extra `-civiform` on everything?

Instead of forking it and keeping the module in a separate git repo can you define the module locally like we do with our other custom modules https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/aws/modules/ ?

cloud/aws/bin/deploy.py
```diff
@@ -18,3 +18,8 @@ def run(config):
 
     aws = AwsCli(config)
     aws.wait_for_ecs_service_healthy()
+    lb_dns = aws.get_load_balancer_dns(f'{config.app_prefix}-civiform-lb')
+    base_url = config.get_base_url()
+    print(
+        f'Server is available at {lb_dns}. Ensure your CNAME record for {base_url} is correct.'
```
bion, line 24: **thought (non-blocking):** Is there any more info we can provide here for the person deploying on what it means for the CNAME to be correct?

cloud/aws/modules/ecs_fargate_service/main.tf
```diff
@@ -0,0 +1,312 @@
...
+# APPLICATION LOAD BALANCER
+#------------------------------------------------------------------------------
+resource "aws_lb" "civiform_lb" {
+  name = substr("${local.name_prefix}-lb", 0, 31)
+
+  internal                         = false
+  load_balancer_type               = "application"
+  drop_invalid_header_fields       = false
+  subnets                          = var.public_subnets
+  idle_timeout                     = 60
```
bion, line 33: **thought (non-blocking):** will be good to be able to parameterize this in the future. cc @dkatzz since it may be relevant for performance work

cloud/aws/modules/ecs_fargate_service/main.tf
```diff
@@ -0,0 +1,312 @@
...
+  to_port           = 443
+  protocol          = "tcp"
+  cidr_blocks       = ["0.0.0.0/0"]
+  prefix_list_ids   = []
+}
+
+moved {
+  from = module.ecs-alb[0].aws_security_group_rule.ingress_through_https["default_http"]
+  to   = aws_security_group_rule.ingress_through_https
+}
```
bion, line 109: **question (non-blocking):** how long do we need to leave this `moved` statements in? until everyone has deployed with these changes at least once?

cloud/aws/modules/ecs_fargate_service/main.tf
```diff
@@ -0,0 +1,312 @@
...
+  min_cpu_period            = var.min_cpu_period
+  scale_target_max_capacity = var.scale_target_max_capacity
+  scale_target_min_capacity = var.scale_target_min_capacity
+  tags                      = var.tags
+}
+
+moved {
+  from = module.ecs-autoscaling[0]
+  to   = module.ecs-autoscaling
+}
```
bion, line 312: **nitpick:** add newline char to end of file

cloud/aws/modules/ecs_fargate_service/outputs.tf
```diff
@@ -0,0 +1,13 @@
...
+}
+output "aws_lb_civiform_lb_arn" {
+  description = "The ARN of the load balancer (matches id)."
+  value       = aws_lb.civiform_lb.arn
+}
+
+output "aws_security_group_lb_access_sg_id" {
+  description = "The ID of the security group"
+  value       = aws_security_group.lb_access_sg.id
+}
```
bion, line 13: **nitpick:** add newline char to end of file

cloud/aws/modules/ecs_fargate_service/variables.tf
```diff
@@ -0,0 +1,134 @@
+#------------------------------------------------------------------------------
+# Misc
+#------------------------------------------------------------------------------
+variable "app_prefix" {
+  description = "App prefix for naming resources on AWS. For most resources, -civiform is added to this string."
```
bion, line 5: **nitpick:** 
```suggestion
  description = "App prefix for naming resources on AWS. For most resources, -civiform is appended to this string."
```

cloud/aws/modules/ecs_fargate_service/variables.tf
```diff
@@ -0,0 +1,134 @@
...
+# AWS ECS SERVICE
+#------------------------------------------------------------------------------
+variable "ecs_cluster_arn" {
+  description = "ARN of an ECS cluster"
+}
+
+variable "desired_count" {
+  description = "(Optional) The number of instances of the task definition to place and keep running. Defaults to 0."
+  type        = number
+  default     = 1
```
bion, line 25: **suggestion (blocking):** 
```suggestion
  description = "(Optional) The number of instances of the task definition to place and keep running. Defaults to 2."
  type        = number
  default     = 2
```

cloud/aws/modules/ecs_fargate_service/variables.tf
```diff
@@ -0,0 +1,134 @@
...
+  description = "(Optional) The name of the SSL Policy for the listener. . Required if var.https_ports is set."
+  type        = string
+  default     = null
+}
+
+variable "default_certificate_arn" {
+  description = "(Optional) The ARN of the default SSL server certificate. Required if var.https_ports is set."
+  type        = string
+  default     = null
+}
```
bion, line 134: **nitpick:** add newline char to end of file

cloud/aws/templates/aws_oidc/app.tf
```diff
@@ -427,3 +398,8 @@ resource "aws_lb_listener_rule" "block_external_traffic_to_metrics_rule" {
     }
   }
 }
+
+moved {
+  from = aws_lb_listener_rule.block_external_traffic_to_metrics_rule[0]
+  to   = aws_lb_listener_rule.block_external_traffic_to_metrics_rule
+}
```
bion, line 405: **nitpick:** add newline char to end of file

bion (approved):

cloud/aws/modules/ecs_fargate_service/variables.tf
bion: > Do you know if we have any existing deployments that set this to 2? Wondering if that's ever been tested.

Good question. I'd thought that all deployments had at least two server nodes running so this is coming as a surprise. I know Seattle does with their old config.

cloud/aws/modules/ecs_fargate_service/main.tf
bion: cool

cloud/aws/modules/ecs_fargate_service/variables.tf
bion: I'm ok merging this as-is if this is at parity with the current system, but definitely something to improve soon to ensure server redundancy in prod envs

bion (approved):
@nb1701 approved, before merging please confirm that this has been vetted by running against an existing deployment

rockycodes (approved):
I don't have a totally clear understanding of what all this is doing, but lgtm from what I understand! Thank you for the helpful comments 🙏

PR #203, opened 2023-06-17
add option to pass in snapshot tag for server var autogeneration

rockycodes:
This PR accomplishes two things:
1. Fixes a previous issue with server var autogeneration where "latest" was getting translated into the latest release version when pulling in the parser package and `env-var-docs.json`, rather than the Docker image tagged "latest"
2. Adds back in the ability to set a specific Docker SNAPSHOT tag as the value for `CIVIFORM_VERSION`

Update: notes below are outdated. For number 1, I implemented the solution where the specific snapshot tag that was used to build the image tagged "latest" is fetched from Docker.

..........................

For number 1, I'm interested in feedback on my approach. When `CIVIFORM_VERSION` is set to "latest", the logic introduced here pulls in the parser package and `env-var-docs.json` as of the last commit to main. This is not a perfect solution, however, because it's possible that the commit tagged "latest" in Docker hub might not be the commit at the tip of main due to a slight delay between merging a commit to main and the image being built in Docker. Moreover, it's possible tests could fail on main and the image would not be built in Docker (although my understanding is this would be because of a flaky test and not an actual test failure since PRs must pass the test suite before being merged to main). 

I see this being an issue in the case that a change is made to the parser package or `env-var-docs.json` that is merged to main but not built in Docker/tagged "latest" for some reason. If a person deploys using the tag "latest" (only used in development and Staging), they would have a newer version of the parser package and/or `env-var-docs.json` than they do of the server code. This could present an issue if, for example, variables were removed from `env-var-docs.json` that the server code relied on.

To solve this "latest" issue, I also explored the option of pulling in the actual `SNAPSHOT` tag from Docker that matches the image tagged "latest" (similar to how we do in [this script](https://github.com/civiform/civiform-staging-deploy/blob/main/bin/resolve-latest-snapshot-tag) on Staging). This would be more precise, but it also adds complexity, a longer deploy time, and additional potential points of failure. I opted for the simpler approach since "latest" is never used in production and (I think) the likelihood of it causing actual problems on Staging or in development is pretty low, but I'm curious what reviewers think!

### DIFF
--- a/cloud/shared/bin/lib/config_loader.py
+++ b/cloud/shared/bin/lib/config_loader.py
@@ -8,6 +8,7 @@ import typing
 import urllib.error
 import urllib.request
 from typing import List
+from typing import Optional
 
 from cloud.shared.bin.lib.config_parser import ConfigParser
 from cloud.shared.bin.lib.print import print
@@ -216,7 +217,7 @@ class ConfigLoader:
 
         return validation_errors
 
-    def _get_commit_sha_for_tag(self, tag: str) -> str | None:
+    def _get_commit_sha_for_tag(self, tag: str) -> Optional[str]:
         """Get the commit SHA for the release specified in the tag.
         
           The tag can be a release version number such as "v1.24.0", a specific docker snapshot 
@@ -245,7 +246,7 @@ class ConfigLoader:
             print(e)
             return None
 
-    def _fetch_json_val(self, url, field_one, field_two=None) -> str | None:
+    def _fetch_json_val(self, url, field_one, field_two=None) -> Optional[str]:
         print(f"Fetching json from url {url}.")
         response = requests.get(url)
 
@@ -257,7 +258,7 @@ class ConfigLoader:
                 f"Error: could not resolve json at {url}. {response.status_code} - {response.json()['message']}"
             )
 
-    def _apply_json_fields(self, json, field_one, field_two) -> str:
+    def _apply_json_fields(self, json, field_one, field_two) -> Optional[str]:
         error_string = f"Error parsing json with fields [{field_one}]"
 
         try:


cloud/shared/bin/lib/config_loader.py
```diff
@@ -214,50 +215,52 @@ def _validate_infra_variables(
...
+            elif "SNAPSHOT" in tag:
+                short_sha = tag.split("-")[1]
+                return self._fetch_json_val(
+                    f"https://api.github.com/repos/civiform/civiform/commits/{short_sha}",
+                    "sha")
+            else:
+                tag_url = self._fetch_json_val(
+                    f"https://api.github.com/repos/civiform/civiform/git/refs/tags/{tag}",
+                    "object", "url")
+                return self._fetch_json_val(tag_url, "object", "sha")
```
bion, line 246: **suggestion (blocking):** please log which of these branches is executes and intermediate values such as `tag_url` on L243 so that if any of them fails we know which it was

cloud/shared/bin/lib/config_loader.py
```diff
@@ -214,50 +215,52 @@ def _validate_infra_variables(
...
+        except self.VersionNotFoundError as e:
+            print(e)
+            return None
```
bion, line 249: **suggestion (blocking):** looks like the return type should be `str | None`

cloud/shared/bin/lib/config_loader.py
```diff
@@ -214,50 +215,52 @@ def _validate_infra_variables(
...
-    def _get_commit_sha_for_tag(self, tag_commit_sha: str) -> str:
-        tag_url = f"https://api.github.com/repos/civiform/civiform/git/tags/{tag_commit_sha}"
-        tag_response = requests.get(tag_url)
-        if tag_response.status_code == 200:
-            commit_sha = tag_response.json()["object"]["sha"]
-            return commit_sha
+            try:
+                return response.json(
+                )[field_one] if field_two is None else response.json(
+                )[field_one][field_two]
```
bion, line 257: **suggestion (blocking):** this is a little tricky to read, please rewrite to avoid having a conditional statement nested inside another statement
https://google.github.io/styleguide/pyguide.html#211-conditional-expressions

cloud/shared/bin/lib/config_loader.py
```diff
@@ -214,50 +215,52 @@ def _validate_infra_variables(
...
+            except:
+                print(f"Error parsing json with fields {field_one} {field_two}")
```
bion, line 259: **suggestion (blocking):** log the json contents as well after the fields

cloud/shared/bin/run
```diff
@@ -40,20 +41,24 @@ function fetch_json_val() {
...
-    "['tag_name']")
-fi
+commit_sha=""
 
-# Get the commit sha for the commit at the tip of the CiviForm version being deployed
-tag_url=$(fetch_json_val \
-  "https://api.github.com/repos/civiform/civiform/git/refs/tags/${version}" \
-  "['object']['url']")
-commit_sha=$(fetch_json_val ${tag_url} "['object']['sha']")
+if [[ ${tag} == "latest" ]]; then
```
bion, line 46: **nitpick (blocking):** quote variables here and L49 https://google.github.io/styleguide/shellguide.html#quoting
```suggestion
if [[ "${tag}" == "latest" ]]; then
```

bion (changes requested):
> This would be more precise, but it also adds complexity, a longer deploy time, and additional potential points of failure.

How much longer would the deploy take? I'd expect a few seconds maybe at most? IMO it'd be worth the complexity/runtime penalties to have fewer ways the deploy can fail. When (not if) it fails due to a mismatch between `latest` the git tag and `latest` the Docker tag, it will likely be someone who is either unfamiliar with (most of the team) or has forgotten (me) the subtlety here.

cloud/shared/bin/lib/config_loader.py
rockycodes: done

cloud/shared/bin/lib/config_loader.py
rockycodes: done

cloud/shared/bin/lib/config_loader.py
rockycodes: done? lmk your thoughts.

cloud/shared/bin/lib/config_loader.py
rockycodes: done

cloud/shared/bin/run
rockycodes: done

rockycodes:
> > This would be more precise, but it also adds complexity, a longer deploy time, and additional potential points of failure.
> 
> How much longer would the deploy take? I'd expect a few seconds maybe at most? IMO it'd be worth the complexity/runtime penalties to have fewer ways the deploy can fail. When (not if) it fails due to a mismatch between `latest` the git tag and `latest` the Docker tag, it will likely be someone who is either unfamiliar with (most of the team) or has forgotten (me) the subtlety here.

@bion Great points. I changed it to resolve the "latest" tag to a specific snapshot tag in `bin/run`. This involved updating some the internals for how "version" was getting checked/used later when fetching `env-var-docs.json` as well. Let me know your thoughts.

cloud/shared/bin/lib/config_loader.py
bion: better, thanks!

cloud/shared/bin/run
```diff
@@ -13,6 +13,8 @@
 # install the package and activate the venv. The activation has to happen in the
 # parent process before the python code is run.
 # This is why we need a bash script that preceeds the python code
+set -e
+set -o pipefail
```
bion, line 17: **suggestion (non-blocking):** we should move toward using the same pattern in these shell scripts that we do in the main repo `bin/` scripts wherein each script begins with `source lib.sh` -- in this case it'd be `source cloud/shared/bin/lib.sh` which sets standard options like these and loads shared code e.g. below where you do `echo "Latest snapshot tag not found." 2>&1` could be `out::error "Latest snapshot tag not found."`

That's a little off-topic for this PR so feel free to merge as-is.

bion (approved):

cloud/shared/bin/run
rockycodes: oo good point! i'll update this.

rockycodes: I ran into some problems refactoring the shell scripts so I'm leaving this as is for now and would love to pair on this with you next week sometime!

PR #206, opened 2023-06-20
Add threshold variables for RDS alarms

dkatzz:
Add additional configuration variables for AWS RDS alarms.

Related to: https://github.com/civiform/civiform/issues/5109

### DIFF
--- a/cloud/aws/templates/aws_oidc/main.tf
+++ b/cloud/aws/templates/aws_oidc/main.tf
@@ -64,28 +64,29 @@ data "aws_db_instance" "civiform" {
 }
 
 module "aws-rds-alarms" {
-  source                                    = "lorenzoaiello/rds-alarms/aws"
-  version                                   = "2.2.0"
-  db_instance_id                            = data.aws_db_instance.civiform.id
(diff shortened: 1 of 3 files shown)


bion (approved):
thought (non-blocking): I wonder if there are helpful consistency validations we could add for numeric config values such as min, max, or step (if the value needs to be a multiple of 8 or something like that). We currently have regex or enum based validations for strings: https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/shared/validate_variable_definitions.py#L101-L130

dkatzz:
> thought (non-blocking): I wonder if there are helpful consistency validations we could add for numeric config values such as min, max, or step (if the value needs to be a multiple of 8 or something like that). We currently have regex or enum based validations for strings: https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/shared/validate_variable_definitions.py#L101-L130

Yes, I like that idea! Added to [doc](https://docs.google.com/document/d/17fQ063Bde1016zdZIsa777OIKgky4qnvqWgzW7qbeIY/edit?pli=1#heading=h.78mhynz94i5i)

PR #210, opened 2023-06-27
Validations and error messages for run script

bion:
Adds some validations and error handling for the run script. Handles:

- [x] User entering a tag without the "v" prefix by adding it for them
- [x] User entering a tag that is syntactically correct but not pointing to a real release
- [x] User entering version that is syntactically incorrect

Fixes https://github.com/civiform/civiform/issues/5140


rockycodes (approved):
this all looks great! thank you for adding it. before merging, please let me know how you tested it to make sure the deploy still works as expected and that all the error paths work as expected.

bion:
@rockycodes yep, ran it locally and observed all code paths. Let's hold of on merging until the current deploy turbulence is resolved

PR #219, opened 2023-07-25
Add APPLICANT_REGISTER_URI into e2e test config

rockycodes:
e2e tests are currently broken because `APPLICANT_REGISTER_URI` is missing from the config: https://github.com/civiform/cloud-deploy-infra/actions/runs/5652272897/job/15311631397

### DIFF
--- a/e2e-test/civiform_config_aws_oidc.sh
+++ b/e2e-test/civiform_config_aws_oidc.sh
@@ -231,6 +231,11 @@ export APPLICANT_OIDC_PROVIDER_NAME="OidcClient"
 # Usually ends in .well-known/openid-configuration
 export APPLICANT_OIDC_DISCOVERY_URI="https://civiform-staging.us.auth0.com/.well-known/openid-configuration"
 
+# REQUIRED if CIVIFORM_APPLICANT_IDP="generic-oidc"
+# The URL applicants are redirected to for creating an account
+# with the identity provider.
+export APPLICANT_REGISTER_URI=""
+
 # OPTIONAL
 # The type of OIDC flow to execute, and how the data is encoded.
 # See https://auth0.com/docs/authenticate/protocols/oauth#authorization-endpoint


e2e-test/civiform_config_aws_oidc.sh
```diff
@@ -231,6 +231,11 @@ export APPLICANT_OIDC_PROVIDER_NAME="OidcClient"
 # Usually ends in .well-known/openid-configuration
 export APPLICANT_OIDC_DISCOVERY_URI="https://civiform-staging.us.auth0.com/.well-known/openid-configuration"
 
+# REQUIRED if CIVIFORM_APPLICANT_IDP="generic-oidc"
+# The URL applicants are redirected to for creating an account
+# with the identity provider.
+export APPLICANT_REGISTER_URI=""
```
bion, line 237: **suggestion (blocking):** can this be set to the auth0 account registration URL?

bion (changes requested):

e2e-test/civiform_config_aws_oidc.sh
rockycodes, line 237:
```suggestion
export APPLICANT_REGISTER_URI="https://staging-aws.civiform.dev/loginForm"
```

rockycodes: Do you know where I would find that? I looked in our auth0 account, but we only have urls set for login and logout.

bion: Looks like it's the same as the login URL but with `screen_hint=signup` added to the params

https://community.auth0.com/t/how-do-i-redirect-users-directly-to-the-hosted-signup-page/42520

rockycodes: chatted with bion about this and we're going to add this value in staging so it's testable.

bion (approved):

PR #229, opened 2023-08-16
Turn on the service deployment circuit breaker

nb1701:
By turning on the circuit breaker, we prevent a deployment with the service crash-looping from repeatedly pulling images from DockerHub, which can cause rate limiting. The deployment will attempt to start the service 10 times. After failing 10 times, it will stop trying and mark the deployment as failed.  Note that this 10 times is a hard-coded minimum limit in AWS.

Additionally, we will now roll back to the previous, working deployment if the new deployment fails. Some logic has been added to detect when this happens, as it will simply appear that the PRIMARY deployment is COMPLETE, and we need to check the deployment ID to detect it's the old deployment.

This also modifies the aws_cli script to detect when it has failed and stops polling the state of the service and prints information on how to check the logs.

Addresses https://github.com/civiform/civiform/issues/4767 and https://github.com/civiform/civiform/issues/5300

### DIFF
--- a/cloud/aws/templates/aws_oidc/bin/aws_cli.py
+++ b/cloud/aws/templates/aws_oidc/bin/aws_cli.py
@@ -55,8 +55,8 @@ class AwsCli:
         """
         Polls the CiviForm ECS service, waiting for the PRIMARY deployment to
         have rolloutStatus of COMPLETED. If the PRIMARY deployment ID ends up
-        being equal to the previous_deployment_id, then the deployment failed
-        and we've rolled back.
+        being different than the ID we attempting to deploy, then the deployment
+        failed and we've rolled back.
 
         Gives up after 60 tries, sleeps 30 seconds between each try.
         """
@@ -71,16 +71,16 @@ class AwsCli:
             For debugging help, contact the CiviForm oncall: https://docs.civiform.us/governance-and-management/project-management/on-call-guide#on-call-responsibilities
             """)
 
-        tracked_deployment_id = None
+        current_deployment_id = None
         tries = 60
         while True:
             info = self._ecs_service_state()
             id = info["id"]
             state = info["state"]
-            tracked_deployment_id = id if tracked_deployment_id is None else tracked_deployment_id
+            current_deployment_id = id if current_deployment_id is None else current_deployment_id
 
             if state == "COMPLETED":
-                if id != tracked_deployment_id:
+                if id != current_deployment_id:
                     print(
                         "ERROR: The deployment that is now COMPLETED has a different ID than the one we were waiting for.\n"
                         "This probably means the new tasks are crash-looping and we've rolled back to the previous deployment.\n"


dkatzz:
If the new deployment fails, would it still use the old task even if rollback is turned off? When I was looking at Arkansas failures, we saw that a new task kept getting created then destroyed but the old task was still running in the meantime. I'm not totally sure if things were still working in that state, but just curious if you tried with rollback off and then had a successful deployment then an unsuccessful deployment what would happen? This approach makes sense to me though if that isn't the case.

rockycodes:
I was gonna chime in with the same question/comment Danielle had. From my experience, the task actually powering CiviForm doesn't switch over to `ACTIVE` _until_ the new deployment has succeeded. This has been working fine with failed deployments, so I'm inclined to leave off the rollback functionality for now and just focus on letting folks know that their deployment failed after 10 tries.

cloud/aws/modules/ecs_fargate_service/main.tf
```diff
@@ -236,9 +236,15 @@ resource "aws_ecs_service" "service" {
     assign_public_ip = false
   }
 
+  # The circuit breaker will mark the deployment as failed when the service does
+  # not come up or is not marked as healthy after 10 tries. This prevents the
+  # deployment from pulling the civiform image over and over, which can cause
+  # DockerHub rate limiting. The 10 tries is the minimum value given that
+  # desired_count < 20 (desired_count is probably 1 for our use case).
```
rockycodes, line 243: question (non-blocking): Just making sure I understand -- It's set to 10 because the desired task count is only 1?

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -50,32 +51,70 @@ def restart_ecs_service(self):
...
+                print(
+                    "ERROR: It appears that the deployment has failed and we've rolled back to the previous working deployment.\n"
+                    "This usually means the new tasks are crash-looping and we've rolled back to the previous deployment.\n"
+                    "To view the logs to see what happened, " + error_text)
+                raise Exception("deployment rolled back")
+
             if state == "COMPLETED":
+                # If the ID of the deployment that is COMPLETED is not the one we've been
+                # waiting for, then:
+                #   - This is a deployment that bypassed the "deployment is already in progress"
```
rockycodes, line 91: question (non-blocking): This list is also confusing to me, but I can't tell if it's because I'm confused by the wording or by the possible scenarios. If this stays in, let's talk through it together.

rockycodes (commented):
Just commenting for now since this might change if you remove the rollback functionality. Overall this looks good to me, but I'm finding the potentially overlapping deploys/many possible states confusing. Let's talk through it if that stays in!

cloud/aws/modules/ecs_fargate_service/main.tf
```diff
@@ -236,9 +236,15 @@ resource "aws_ecs_service" "service" {
...
+  # https://docs.aws.amazon.com/AmazonECS/latest/userguide/deployment-circuit-breaker.html
   deployment_circuit_breaker {
-    enable   = false
-    rollback = false
+    enable   = true
+    rollback = true
```
dkatzz, line 247: **thought:** I'm still thinking about whether it makes sense to turn on this rollback functionality if it works with it off. With the last deployment issue, it would sometimes work and other times wouldn't (regardless of the version), since there was a race condition. So I almost think it may be safer to keep rollback functionality off to use the deployment that was successful in the case that we can't create another successful deployment. Thoughts?

bion (approved):

cloud/aws/modules/ecs_fargate_service/main.tf
dkatzz: Ohh that makes sense! Thanks!

dkatzz (approved):

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -53,47 +54,70 @@ def restart_ecs_service(self):
...
         tries = 60
         while True:
-            state = self._ecs_service_state()
+            info = self._ecs_service_state()
+            id = info["id"]
+            state = info["state"]
+            tracked_deployment_id = id if tracked_deployment_id is None else tracked_deployment_id
+
             if state == "COMPLETED":
+                if id != tracked_deployment_id:
```
rockycodes, line 83: suggestion (non-blocking): IIUC, `id` here would be the id of the deployment we rolled back to, and `tracked_deployment_id` is the id of the deployment we were attempting? What do you think about renaming `tracked_deployment_id` to `current_deployment_id`? That feels more clear to me, but also clarity is subjective.

rockycodes (approved):
This looks great and is much clearer! I left a couple of optional comments that I think might further help with clarity, but feel free to ignore/resolve them if you disagree.

PR #235, opened 2023-08-22
Add ability to force unlock TF state and fix digest value of lock file in DynamoDB

nb1701:
Sometimes, if a failure occurs or you've Ctrl+C'ed out of the middle of an apply, the Terraform state may be left in a locked state. When this happens, you must run 'terraform force-unlock' with the lock ID to unlock it.

This change detects when an apply results in a lock error, and prompts the user to rerun the command with the --force-unlock flag. It does this by capturing the stderr stream of the 'terraform apply' command to inspect the message if an error occurs. An added bonus is that since we are using Popen to execute the command, we can handle things more gracefully if a user Ctrl+Cs out of the apply and allow Terraform to release the lock, reducing the amount of ways we can get into this failure state.

Secondly, if something goes wrong during deployment, especially when a user has force-unlocked due to a previous issue and then multiple apply actions are happening at once, the digest value for the Terraform lock file in S3 can be incorrect. This lets us set the digest value to the correct value, as given by the error message of a previous Terraform command, without having to go into the AWS console to set it manually.

Requires https://github.com/civiform/civiform-deploy/pull/50 and addresses https://github.com/civiform/civiform/issues/5381

This also adds a `bin/fmt` script to format files like the checks are expecting. It is not containerized and requires the binaries to be installed locally, so we probably want to containerize it later.  But it works as a first pass.

### DIFF
--- a/cloud/shared/bin/lib/terraform.py
+++ b/cloud/shared/bin/lib/terraform.py
@@ -51,6 +51,7 @@ def perform_init(
     print(f" - Run {init_cmd}")
     output, exit_code = capture_stderr(init_cmd)
     if exit_code > 0:
+        # Determine if we're running interactively
         is_tty = sys.stdin.isatty()
         # This is AWS-specific, and should be modified when we have actual
         # Azure deployments
@@ -69,7 +70,7 @@ def perform_init(
                             config_loader, terraform_template_dir, upgrade)
                         return
                 print(
-                    f"To fix the above error, set the LOCK_TABLE_DIGEST_VALUE environment variable to \"{digest}\" and rerun this command."
+                    f"To fix the above error, rerun the command with LOCK_TABLE_DIGEST_VALUE=\"{digest}\" before it."
                 )
             # Since we've handled the error and printed a message, exit immediately
             # rather than returning False and having it print a stack trace.
@@ -139,6 +140,7 @@ def perform_apply(
 
     output, exit_code = capture_stderr(terraform_apply_cmd)
     if exit_code > 0:
+        # Determine if we're running interactively
         is_tty = sys.stdin.isatty()
         if "Error acquiring the state lock" in output:
             # Lock ID is a standard UUID v4 in the form 00000000-0000-0000-0000-000000000000
@@ -165,13 +167,13 @@ def perform_apply(
                             False)  # initialize = False
                 print(
                     error_text +
-                    f"\nIf you are sure there are no other Terraform processes running, this can be fixed by setting the FORCE_UNLOCK_ID environment variable to \"{lock_id}\" and rerunning the same command."
+                    f"\nIf you are sure there are no other Terraform processes running, this can be fixed by rerunning the command with FORCE_UNLOCK_ID=\"{lock_id}\" before it."
                 )
             else:
                 print(
                     error_text +
                     "\nWe were unable to extract the lock ID from the error text. Inspect the error message above."
-                    "\nIf you are sure there are no other Terraform processes running, this error can be fixed by setting the FORCE_UNLOCK_ID environment variable to the lock ID value, and then rerunning the same command."
+                    "\nIf you are sure there are no other Terraform processes running, this error can be fixed by rerunning the command with FORCE_UNLOCK_ID=<Lock ID> before it."
                 )
             # Since we've handled the error and printed a message, exit immediately
             # rather than returning False and having it print a stack trace.


bin/fmt
dkatzz, line 1: **praise:** Thanks for adding this!!

cloud/shared/bin/lib/terraform.py
```diff
@@ -1,37 +1,97 @@
...
+        universal_newlines=True)
+    try:
+        exit_code = popen.wait()
+        _, stderr = popen.communicate()
+        if stderr:
+            print(stderr)
+        return stderr, exit_code
+    except KeyboardInterrupt:
+        # Allow terraform to gracefully exit if a user Ctrl+C's out of the command
+        #popen.terminate()
```
dkatzz, line 80: **nitpick:** Should be deleted?

cloud/shared/bin/lib/terraform.py
```diff
@@ -1,37 +1,97 @@
...
-        print(' - Run terraform init -upgrade')
-        init_cmd = f'{terraform_cmd} init -input=false -upgrade'
+        init_cmd += ' -input=false'
         # backend vars file can be absent when pre-terraform setup is running
         if os.path.exists(os.path.join(terraform_template_dir,
                                        config_loader.backend_vars_filename)):
             init_cmd += f' -backend-config={config_loader.backend_vars_filename}'
-        subprocess.check_call(shlex.split(init_cmd))
+    print(f" - Run {init_cmd}")
+    #subprocess.check_call(shlex.split(init_cmd))
```
dkatzz, line 48: **nitpick:**  Should be deleted?

dkatzz (approved):

cloud/shared/bin/lib/terraform.py
```diff
@@ -45,16 +103,42 @@ def perform_apply(
...
+            # Lock ID is a standard UUID v4 in the form 00000000-0000-0000-0000-000000000000
+            match = re.search(
+                    r'ID:\s+([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})',
+                    output)
+            error_text = inspect.cleandoc(
+                """
+                                     The Terraform state lock can not be acquired.
+                                     This can happen if you are running a command in another process, or if another Terraform process exited prematurely. 
+                                     """)
+            if match:
```
dkatzz, line 129: **thought:** If there is a match, is there a way of raising an error and prompting the user if they want to force unlock with the matched ID and if they do we do it for them without them having to re-run? 

Just thinking that would allow civic entities to proceed without needing to update civiform-deploy hypothetically, so may be good to have if possible.

dkatzz: Got it, yeah that makes sense! I think giving the option of doing it for them if they're running interactively already is nice and printing the message if not SGTM. I'll write in the thread as well.

bin/fmt
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
bion, line 1: **nitpick:** 
```suggestion
#! /usr/bin/env bash
```

bin/fmt
```diff
@@ -0,0 +1,29 @@
...
+
+# DOC: Format terraform, python, and shell script files.
+
+if which -s terraform; then
+  echo "Formatting Terraform files"
+  terraform fmt -recursive -write
+else
+  echo "Terraform not found. Skipping."
```
bion, line 9: **suggestion (blocking):** should exit with an error code if it can't perform its function

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -109,6 +109,23 @@ def wait_for_ecs_service_healthy(self):
             )
             time.sleep(30)
 
+    def fix_digest_value(self, value):
+        """
+        Sets the lock file digest value in DynamoDB to the given value.
```
bion, line 114: **suggestion (blocking):** would be helpful to mention that the digest is a chucksum of the terraform state file

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -109,6 +109,23 @@ def wait_for_ecs_service_healthy(self):
             )
             time.sleep(30)
 
+    def fix_digest_value(self, value):
```
bion, line 112: **suggestion (blocking):** this method is a bit more generic than "fix" since it really just sets the value, recommend renaming it to `set_lock_table_digest_value` and keeping the discussion of why we do this in the pydoc

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -109,6 +109,23 @@ def wait_for_ecs_service_healthy(self):
...
+        are happening at once, the digest value for the Terraform lock file in
+        S3 can be incorrect. This function lets us set the digest value to
+        the correct value, as given by the error message of a previous
+        Terraform command, without having to go into the AWS console to
+        set it manually.
+        """
+        table = f'{self.config.app_prefix}-{resources.S3_TERRAFORM_LOCK_TABLE}'
+        file = f'{self.config.app_prefix}-{resources.S3_TERRAFORM_STATE_BUCKET}'
+        command = f'dynamodb put-item --table-name={table} --item=\'{{"LockID":{{"S":"{file}/tfstate/terraform.tfstate-md5"}},"Digest":{{"S":"{value}"}}}}\''
+        self._call_cli(command, False)
```
bion, line 127: **nitpick:** either add an inline comment hinting at what the boolean is or use an enum param instead of a boolean

cloud/shared/bin/lib/terraform.py
```diff
@@ -1,37 +1,95 @@
...
         terraform_template_dir = config_loader.get_template_dir()
-    tf_vars_filename = config_loader.tfvars_filename
 
-    terraform_cmd = f'terraform -chdir={terraform_template_dir}'
+    perform_init(config_loader, terraform_template_dir, False)
+
+    terraform_cmd = f'terraform -chdir={terraform_template_dir} force-unlock -force {lock_id}'
+    print(f" - Run {terraform_cmd}")
+    subprocess.check_call(shlex.split(terraform_cmd))
+    return True
```
bion, line 25: **suggestion:** doesn't look like this return value is consumed
```suggestion
```

cloud/shared/bin/lib/terraform.py
```diff
@@ -1,37 +1,95 @@
...
-        is_destroy=False,
+        lock_id: str,
         terraform_template_dir: Optional[str] = None):
-    '''Generates terraform variable files and runs terraform init and apply.'''
     if not terraform_template_dir:
         terraform_template_dir = config_loader.get_template_dir()
-    tf_vars_filename = config_loader.tfvars_filename
 
-    terraform_cmd = f'terraform -chdir={terraform_template_dir}'
+    perform_init(config_loader, terraform_template_dir, False)
```
bion, line 20: **nitpick:** either add an inline comment hinting at what the boolean is or use an enum param instead of a boolean

cloud/shared/bin/lib/terraform.py
```diff
@@ -1,37 +1,95 @@
...
+    if exit_code:
+        # This is AWS-specific, and should be modified when we have actual
+        # Azure deployments
+        if 'state data in S3 does not have the expected content' in output:
+            match = re.search(r'value: ([0-9a-f]{32})', output)
+            if match:
+                print(
+                    f"To fix the above error, rerun this command with \"--fix-digest={match.group(match.lastindex)}\""
+                )
+            exit(exit_code)
```
bion, line 58: **suggestion (blocking):** let's not fail silently for other cloud providers. Use `config_loader.get_cloud_provider()` to ensure the provider is AWS, throw `NotImplementedError` if it's not

bion (changes requested):

cloud/shared/bin/lib/terraform.py
```diff
@@ -1,37 +1,118 @@
...
+        init_cmd += ' -input=false'
         # backend vars file can be absent when pre-terraform setup is running
         if os.path.exists(os.path.join(terraform_template_dir,
                                        config_loader.backend_vars_filename)):
             init_cmd += f' -backend-config={config_loader.backend_vars_filename}'
-        subprocess.check_call(shlex.split(init_cmd))
+    print(f" - Run {init_cmd}")
+    output, exit_code = capture_stderr(init_cmd)
+    if exit_code > 0:
+        is_tty = sys.stdin.isatty()
```
dkatzz, line 54: **nitpick:** Maybe add a comment above saying this is checking for interactive mode?

cloud/shared/bin/lib/terraform.py
```diff
@@ -1,37 +1,118 @@
...
+                        "Would you like to fix this by setting the correct digest value? Ensure that no other deployment processes are in progress. [Y/n] >"
+                    )
+                    if answer.lower() in ['y', 'yes', '']:
+                        aws = AwsCli(config_loader)
+                        aws.set_lock_table_digest_value(digest)
+                        perform_init(
+                            config_loader, terraform_template_dir, upgrade)
+                        return
+                print(
+                    f"To fix the above error, set the LOCK_TABLE_DIGEST_VALUE environment variable to \"{digest}\" and rerun this command."
```
dkatzz, line 72: **suggestion:** You'll need to add that variable to env-var-docs, right? Also, what happens if they add that lock id and then forget to remove it?

dkatzz: Ah got it, thanks! Maybe we can make the print statement more clear by saying something like:

`To fix the above error, rerun the command with LOCK_TABLE_DIGEST_VALUE=\"{digest}\" before it.`

Thoughts? Just so there isn't any confusion about where it belongs.

bion (approved):

PR #237, opened 2023-08-23
Use Python 3.7 for testing

nb1701:
This sets up Python 3.7 to run tests against, since this is the minimum
version we support for the deploy system. This will help us ensure we
are not adding code that only works on >3.7.


rockycodes (approved):
smart!

PR #254, opened 2023-09-12
Update pgadmin version

dkatzz:
Seeing some updates with v7.6, so downgrading while I investigate further.


avaleske (approved):
@avaleske approved this PR from Slack with [Graphite](https://app.graphite.dev/github/pr/civiform/cloud-deploy-infra/254)

PR #256, opened 2023-09-12
Exit script when Terraform apply fails

nb1701:
This checks the result of perform_apply when running bin/setup so that it exits the script appropriately when it fails and doesn't attempt to do post-apply actions. If the apply fails for any reason (including a user entering "no" when asked if they want to do the apply), it exits with a non-zero code. Previously, the subprocess.check_call would have caught this and raised an exception. Now, since we wrap the call in a subprocess.Popen and simply return False if the exit code is non-zero, we have to handle this case differently.

Additionally, this removes the logic where we retry the apply if it fails. This bug in the Terraform aws provider was fixed in v5, which we are currently using.


rockycodes (approved):
lgtm! that's great they fixed the bug, and thanks for cleaning up our related code!

PR #261, opened 2023-09-21
Environmentalize email sns topic

dkatzz:
We should have the app prefix with the sns ID, so there aren't issues with different environments

### DIFF
--- a/cloud/aws/templates/aws_oidc/alarms.tf
+++ b/cloud/aws/templates/aws_oidc/alarms.tf
@@ -1,7 +1,7 @@
 // SNS topic to alert if an alarm gets triggered
 resource "aws_sns_topic" "civiform_alert_topic" {
   count = var.rds_alarm_email != "" ? 1 : 0
-  name  = "civiform-alert-topic"
+  name  = "${var.app_prefix}-civiform-alert-topic"
 }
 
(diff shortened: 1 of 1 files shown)


cloud/aws/templates/aws_oidc/alarms.tf
```diff
@@ -1,7 +1,7 @@
 // SNS topic to alert if an alarm gets triggered
 resource "aws_sns_topic" "civiform_alert_topic" {
   count = var.rds_alarm_email != "" ? 1 : 0
-  name  = "civiform-alert-topic"
+  name  = "${var.app_prefix}-civiform-alert-topic"
```
rockycodes, line 4: Did you test this on an already deployed instance? I'm wondering if renaming it would break anything for existing deploys.

dkatzz: Yeah I did! If you already have one that's deployed it will replace your sns topic for this new one. The issue with the current implementation is if you deploy one environment it will create the SNS topic and then for the second environment it will try to create another SNS topic and fail.

rockycodes (approved):

cloud/aws/templates/aws_oidc/alarms.tf
rockycodes: Thanks for explaining! Sounds good to me 👍

PR #263, opened 2023-10-16
add high memory utilization alarm

elliotgreenlee:
Add an alarm to check if the MemoryUtilization metric is above a threshold (default 80%).

[Cloudwatch metrics](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cloudwatch-metrics.html)

Tested on local deployment

### DIFF
--- a/cloud/aws/templates/aws_oidc/alarms.tf
+++ b/cloud/aws/templates/aws_oidc/alarms.tf
@@ -195,7 +195,7 @@ resource "aws_cloudwatch_metric_alarm" "maximum_used_transaction_ids_too_high" {
 
 resource "aws_cloudwatch_metric_alarm" "memory_utilization_too_high" {
   count               = var.ecs_create_high_memory_alarm ? 1 : 0
-  alarm_name          = "ecs-${module.ecs_cluster.name}-highMemoryUtilization"
+  alarm_name          = "ecs-${module.ecs_cluster.aws_ecs_cluster_cluster_name}-highMemoryUtilization"
   comparison_operator = "GreaterThanThreshold"
   evaluation_periods  = var.ecs_alarm_evaluation_period
   metric_name         = "MemoryUtilization"
@@ -207,6 +207,6 @@ resource "aws_cloudwatch_metric_alarm" "memory_utilization_too_high" {
   alarm_actions       = local.civiform_alarm_actions
 
   dimensions = {
-    ClusterName = module.ecs_cluster.name
+    ClusterName = module.ecs_cluster.aws_ecs_cluster_cluster_name
   }
 }
--- a/cloud/aws/templates/aws_oidc/alarms.tf
+++ b/cloud/aws/templates/aws_oidc/alarms.tf
@@ -34,6 +34,24 @@ resource "aws_cloudwatch_metric_alarm" "cpu_utilization_too_high" {
   }
 }
 
+resource "aws_cloudwatch_metric_alarm" "memory_utilization_too_high" {
+  count               = var.rds_create_high_memory_alarm ? 1 : 0
+  alarm_name          = "rds-${data.aws_db_instance.civiform.id}-highMemoryUtilization"
+  comparison_operator = "GreaterThanThreshold"
+  evaluation_periods  = var.rds_alarm_evaluation_period
+  metric_name         = "MemoryUtilization"
+  namespace           = "AWS/RDS"
+  period              = var.rds_alarm_statistic_period
+  statistic           = "Average"
+  threshold           = var.rds_max_memory_utilization_threshold
+  alarm_description   = "Average database memory utilization is too high."
+  alarm_actions       = local.rds_alarm_actions
+
+  dimensions = {
+    DBInstanceIdentifier = data.aws_db_instance.civiform.id
+  }
+}
+
 resource "aws_cloudwatch_metric_alarm" "cpu_credit_balance_too_low" {
   count               = var.rds_create_low_cpu_credit_alarm ? length(regexall("(t2|t3|t4)", var.postgres_instance_class)) > 0 ? 1 : 0 : 0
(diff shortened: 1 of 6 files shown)


cloud/aws/templates/aws_oidc/alarms.tf
```diff
@@ -34,6 +34,25 @@ resource "aws_cloudwatch_metric_alarm" "cpu_utilization_too_high" {
   }
 }
 
+// Memory Utilization
+resource "aws_cloudwatch_metric_alarm" "memory_utilization_too_high" {
+  count               = var.rds_create_high_memory_alarm ? 1 : 0
+  alarm_name          = "rds-${data.aws_db_instance.civiform.id}-highMemoryUtilization"
```
dkatzz, line 40: **suggestion:** Sorry for not noticing this yesterday, but currently this is set up for RDS, but we'll want it for ECS instead. This will involve updating any RDS references in the variable names and the alert values to ECS.

elliotgreenlee: ok! should I also add a cpu utilization alarm for ecs?

dkatzz: We actually have one setup as part of the [ECS autoscaling policy](https://github.com/civiform/cloud-deploy-infra/blob/40c339725e080607a22708279e320ce656ff3cab/cloud/aws/modules/ecs_fargate_service/main.tf#L297). It currently is set up as part of another [terraform module](https://github.com/cn-terraform/terraform-aws-ecs-service-autoscaling), but we have another issue to remove that dependency, so its more obvious that the alarm exists https://github.com/civiform/civiform/issues/5632. So don't worry about adding another alarm for ECS CPUUtilization, but thank you for thinking of that!

elliotgreenlee: Should I:
* generalize the actions email so it isn't rds_alarm_email
* create a new ecs actions email and sns topic?

dkatzz: Good question! Let's generalize, but Arkansas is using the rds_alarm_email variable already, so we'll have to let them know to update that when they run the next deployment.

cloud/aws/templates/aws_oidc/alarms.tf
```diff
@@ -190,5 +190,24 @@ resource "aws_cloudwatch_metric_alarm" "maximum_used_transaction_ids_too_high" {
...
+  namespace           = "AWS/ECS"
+  period              = var.ecs_alarm_statistic_period
+  statistic           = "Average"
+  threshold           = var.ecs_max_memory_utilization_threshold
+  alarm_description   = "Average ECS service memory utilization is too high."
+  alarm_actions       = local.civiform_alarm_actions
+
+  dimensions = {
+    ClusterName = module.ecs_cluster.aws_ecs_cluster_cluster_name
+    ServiceName = "${module.ecs_cluster.aws_ecs_cluster_cluster_name}-service"
```
elliotgreenlee, line 211: Is there a better way to set this correctly?

dkatzz: Does `aws_ecs_service.service.name` work?

cloud/aws/templates/aws_oidc/alarms.tf
```diff
@@ -190,5 +190,24 @@ resource "aws_cloudwatch_metric_alarm" "maximum_used_transaction_ids_too_high" {
   statistic           = "Average"
   threshold           = var.rds_max_used_transaction_ids_high_threshold
   alarm_description   = "Nearing a possible critical transaction ID wraparound. More info [here](https://aws.amazon.com/blogs/database/implement-an-early-warning-system-for-transaction-id-wraparound-in-amazon-rds-for-postgresql/)"
-  alarm_actions       = local.rds_alarm_actions
+  alarm_actions       = local.civiform_alarm_actions
+}
+
+resource "aws_cloudwatch_metric_alarm" "memory_utilization_too_high" {
+  count               = var.ecs_create_high_memory_alarm ? 1 : 0
+  alarm_name          = "ecs-${module.ecs_cluster.aws_ecs_cluster_cluster_name}-highMemoryUtilization"
```
dkatzz, line 198: **nitpick:** Thoughts on using `local.name_prefix` here?

elliotgreenlee: I have gut reservations about not explicitly dimensioning the name on the dimension criteria, but I am not familiar enough to say why that might cause problems later. I think it's a pretty easy fix so let's do it.

cloud/aws/templates/aws_oidc/alarms.tf
elliotgreenlee: Yes, but I needed to 1. Understand our file structure and general terraform file structures better 2. add that value as an output of the ecs_fargate_service module.

cloud/aws/templates/aws_oidc/variables.tf
```diff
@@ -547,6 +541,36 @@ variable "ecs_min_cpu_period" {
   default     = "60"
 }
 
+variable "ecs_task_memory" {
```
dkatzz, line 544: I think maybe we should leave this where it was previously, so that it lives with the other ECS variables (and the alarm variables stay down here). Thoughts? (For both this file and the file above)

dkatzz (approved):
Just one comment about the placement of the ECS memory variable, but otherwise LGTM! Awesome job!

dkatzz:
Also, when you submit this, please post in the arkansas-deployment channel tagging Terry to let him know that he should rename the `RDS_ALARM_EMAIL` variable to `CIVIFORM_ALARM_EMAIL`

cloud/aws/templates/aws_oidc/variables.tf
elliotgreenlee: ok!

PR #264, opened 2023-10-18
Add variable to disable monitoring stack

bion:
Adds a variable that controls whether or not to create Prometheus and Grafana instances in AWS. This is intended to be used for staging environments where the deployer doesn't want monitoring and would like to save on the expense.

Tested: ran `bin/deploy` for an existing deployment which reported that existing prometheus and grafana resources are moved i.e. not destroyed and replaced:
```
  # aws_grafana_workspace.CiviForm_metrics has moved to aws_grafana_workspace.CiviForm_metrics[0]
    resource "aws_grafana_workspace" "CiviForm_metrics" {
        id                        = "g-866ca756db"
        name                      = "cf-test-civiform-metrics"
        tags                      = {}
        # (13 unchanged attributes hidden)
    }

  # aws_prometheus_workspace.metrics has moved to aws_prometheus_workspace.metrics[0]
    resource "aws_prometheus_workspace" "metrics" {
        id                  = "ws-4caf9bf0-89c9-4047-8239-6cca617f4621"
        tags                = {}
        # (4 unchanged attributes hidden)
    }
```

### DIFF
--- a/cloud/aws/templates/aws_oidc/app.tf
+++ b/cloud/aws/templates/aws_oidc/app.tf
@@ -344,8 +344,8 @@ resource "aws_iam_role_policy_attachment" "civiform_ecs_task_execution_role_cust
   policy_arn = aws_iam_policy.civiform_ecs_task_execution_role_custom_policy[count.index].arn
 }
 
-resource "aws_ecs_task_definition" "td" {
-  family = "${local.name_prefix}-td"
+resource "aws_ecs_task_definition" "civiform_with_monitoring" {
+  family = "${local.name_prefix}-civiform-with-monitoring-td"
 
   cpu    = var.ecs_task_cpu
   memory = var.ecs_task_memory
@@ -362,6 +362,21 @@ resource "aws_ecs_task_definition" "td" {
   tags                     = local.tags
 }
 
+resource "aws_ecs_task_definition" "civiform_only" {
+  family = "${local.name_prefix}-civiform-only-td"
+
+  cpu    = var.ecs_task_cpu
+  memory = var.ecs_task_memory
+
+  container_definitions = jsonencode([module.civiform_server_container_def.json_map_object])
+
+  task_role_arn            = aws_iam_role.civiform_ecs_task_execution_role.arn
+  execution_role_arn       = aws_iam_role.civiform_ecs_task_execution_role.arn
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["FARGATE"]
+  tags                     = local.tags
+}
+
 module "ecs_fargate_service" {
   source                    = "../../modules/ecs_fargate_service"
   app_prefix                = var.app_prefix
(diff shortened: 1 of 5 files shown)


cloud/aws/templates/aws_oidc/app.tf
```diff
@@ -211,7 +211,7 @@ module "civiform_metrics_scraper_container_def" {
   container_memory_reservation = 1024
```
dkatzz, line 212: **thought:** Trying to think of why this wouldn't work. Should we attempt to only build this scraper container if the metrics are enabled? Similar to what we did with the count for other resources?

bion: I tried that and terraform doesn't allow the number of containers in a task def to change based on a variable

dkatzz: Hmm would it work if you updated this container_definitions variable based on that variable? https://github.com/civiform/cloud-deploy-infra/blob/40c339725e080607a22708279e320ce656ff3cab/cloud/aws/templates/aws_oidc/app.tf#L353

bion: That's what I tried updating, Terraform doesn't want that to change. I guess an alternative could be to create multiple task definitions and switch which one we use based on the variable?

cloud/aws/templates/aws_oidc/app.tf
```diff
@@ -344,8 +344,8 @@ resource "aws_iam_role_policy_attachment" "civiform_ecs_task_execution_role_cust
   policy_arn = aws_iam_policy.civiform_ecs_task_execution_role_custom_policy[count.index].arn
 }
 
-resource "aws_ecs_task_definition" "td" {
-  family = "${local.name_prefix}-td"
+resource "aws_ecs_task_definition" "civiform_with_monitoring" {
```
dkatzz, line 347: **thought:** I'm fine with either option but wondering if we should keep the same name for the existing instance (with a comment that this one has monitoring) and update the name for the new instance. Just to ensure it isn't confusing when entities see something new. And the other definition can be "td_without_monitoring" or "civiform_only" or some combo of those.

dkatzz (approved):

cloud/aws/templates/aws_oidc/app.tf
bion: I'd like to move away from inscrutable names like `td_`.

> Just to ensure it isn't confusing when entities see something new.

I don't know how much people are reading the terraform output when they run a deployment. That being said, the only deployer using this system who isn't already aware of this change is Terry, it'd be easy to let him know to expect this. The benefit of the clearer variable name is worth it imo.

dkatzz: That makes sense to me! Would you mind just mentioning in the arkansas-deployment channel once this is submitted.

bion: @dkatzz will do!

PR #265, opened 2023-10-24
Removing phone question type env. variable

swatkat1:
Removing phone question type env. variable


bion (changes requested):
Please clean up the other references as well
image

swatkat1:
Hi @bion I have updated. But can the docs be updated too? None of these docs are mentions for feature flag removal in the docs - https://github.com/civiform/civiform/wiki/Feature-Flags

bion (approved):

PR #268, opened 2023-10-30
Cleanup autogenerated server vars

rockycodes:
This PR removes variables that are generated/referenced by server variable auto-generation now and are no longer needed. It also removes a method (`overwrite_checkout_file`) that was used to rollout server var auto-generation and is no longer needed.

I mostly focused on removing vars from the shared files and AWS since those are our active deployments, but I also removed some vars from Azure that are clearly no longer needed.

The variables removed in this PR are listed [in this doc](https://docs.google.com/spreadsheets/d/1XPPXYzgqutC70FWx7r-8flIzOl7JEOrQ_1I5mNX4lU0/edit?usp=sharing) and grouped by the reason it is safe to remove them. There are 54 of them (!!). I erred on the side of caution when deciding what to remove, but I am not familiar with all of these variables and would appreciate a second eye.

Testing:
New deployments:  I ran `bin/setup` with the changes on this branch and confirmed it deployed successfully and basic functionality worked.

Existing deployments: I ran `bin/setup` using the code on main then `bin/deploy` using the code on this branch and confirmed deployment was successful and nothing obvious broke or changed. I compared the admin settings panel pages and made sure all values were consistent. The only difference is the `ESRI_FIND_ADDRESS_CANDIDATES_URL` which previously defaulted to a value set in the deploy system, but will now be blank unless set in the config file.

After this is merged, I'll test deploying it on staging to ensure login functionality isn't affected.

Closes: https://github.com/civiform/civiform/issues/5249


bion (changes requested):
Thank you!!! This has been a long time coming, great to see it finally happen.

It looks like there are a few vars that could also be removed from here? https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/aws/templates/aws_oidc/variable_definitions.json#L44

rockycodes:
> Thank you!!! This has been a long time coming, great to see it finally happen.
> 
> It looks like there are a few vars that could also be removed from here? https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/aws/templates/aws_oidc/variable_definitions.json#L44

@bion what vars in particular? I'm removing the line you linked to [here](https://github.com/civiform/cloud-deploy-infra/pull/268/files#diff-d5f0de2a0a2dad8f4fe13e04fa34e7cf2389688766f0d656873bfc4c8c25f44fL44)

rockycodes:
> Looks good! I wonder if some of those marked as "used elsewhere" could be refactored later, e.g. AWS_SES_SENDER is really SENDER_EMAIL_ADDRESS, can we just use AWS_SES_SENDER everywhere instead of SENDER_EMAIL_ADDRESS? But definitely not needed for this one.

Good idea, I can make an issue to follow up with that.

bion (approved):

PR #280, opened 2024-01-04
[WIP] Pull/Push Script for Program Data

elliotgreenlee:
Pull program data and push program data without needing to use the pgadmin UI.

Open a pgadmin instance and connect in way that commands can be run on the database from it using existing database protections. Run push or pull command (or in the future, any command). 

Changes:
* Add a command to pull program data ([pg_dump](https://www.postgresql.org/docs/current/app-pgdump.html)) to ./bin/run.
* Add a command to push program data ([pg_restore](https://www.postgresql.org/docs/current/app-pgrestore.html)) to ./bin/run.
*  Add a command to enter an interactive shell to ./bin/run.
* Generalize pgadmin startup code.
* Add [ecs list-tasks](https://docs.aws.amazon.com/cli/latest/reference/ecs/list-tasks.html) to our aws cli.
* Add [ecs execute-command](https://docs.aws.amazon.com/cli/latest/reference/ecs/execute-command.html) to our aws cli.
* Add [rds describe-db-instances](https://docs.aws.amazon.com/cli/latest/reference/rds/describe-db-instances.html) to our aws cli.

I believe it is acceptable to pass the db_password as part of the constructed command rather than using a [.pgpass file](https://www.postgresql.org/docs/current/libpq-pgpass.html), since our password and username and generated randomly and not stored. https://stackoverflow.com/q/2893954 is a good reference for options.


dkatzz:
Chatted with Elliot and they mentioned:

- This PR involves adding some things to the AWS CLI to make that interaction easier
- 90- 283 can swap out any command -> should be the same for whatever you’re doing -> db interactive shell has those lines so doesn’t need to be in every file
- Tested up to line 70 on pg dump
- These commands aren't connected to bin/run, but that could be added later

dkatzz:
This functionality has mostly been added as part of https://github.com/civiform/cloud-deploy-infra/pull/311 and https://github.com/civiform/cloud-deploy-infra/pull/310.

PR #283, opened 2024-01-29
trim aws load balancer name to 32 characters (max size)

elliotgreenlee:
AWS load balancer name can't be longer than 32 characters. https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-application-load-balancer.html#configure-load-balancer

```
│ Error: "name" cannot be longer than 32 characters
│ 
│   with module.pgadmin[0].aws_lb_target_group.pgadmin,
│   on ../../modules/pgadmin/main.tf line 29, in resource "aws_lb_target_group" "pgadmin":
│   29:   name        = local.name_prefix
│ 
```

### DIFF
--- a/cloud/aws/modules/pgadmin/main.tf
+++ b/cloud/aws/modules/pgadmin/main.tf
@@ -26,7 +26,7 @@ resource "aws_lb_listener" "pgadmin" {
 
 # Traffic from load balancer is forwarded to IPs in this target group.
 resource "aws_lb_target_group" "pgadmin" {
-  name        = substr(local.name_prefix, 0, 32)
+  name        = local.name_prefix
   port        = 80
   protocol    = "HTTP"
   target_type = "ip"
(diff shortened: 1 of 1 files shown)


dkatzz (commented):
Thanks for doing this!

cloud/aws/modules/pgadmin/main.tf
```diff
@@ -26,7 +26,7 @@ resource "aws_lb_listener" "pgadmin" {
 
 # Traffic from load balancer is forwarded to IPs in this target group.
 resource "aws_lb_target_group" "pgadmin" {
-  name        = local.name_prefix
+  name        = substr(local.name_prefix, 0, 32)
```
dkatzz, line 29: **suggestion:** Are there other resources that might run into this? Thoughts on updating the prefix local instead? And / or we can change it from "${var.app_prefix}-civiform-pgadmin" to "${var.app_prefix}-cf-pgadmin" to make this less likely to happen https://github.com/civiform/cloud-deploy-infra/blob/91ce382eef8f87804d72d8d1c0a74706b3b86aa5/cloud/aws/modules/pgadmin/main.tf#L99C17-L99C53

elliotgreenlee: oh I should add a comment explaining why

PR #284, opened 2024-01-30
correct pgadmin prefix for secrets to reflect app name

elliotgreenlee:
https://github.com/civiform/cloud-deploy-infra/pull/283 shortened civiform to "cf", but did not correct other uses.

### DIFF
--- a/cloud/aws/bin/pgadmin.py
+++ b/cloud/aws/bin/pgadmin.py
@@ -102,7 +102,7 @@ def _wait_for_pgadmin_response(url):
 
 def _print_connection_info(config, url):
     aws = AwsCli(config)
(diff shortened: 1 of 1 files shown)


elliotgreenlee:
I would rather use the same variable for both things, but my grasp of the variable link between terraform and this python file is a little tenuous.

elliotgreenlee:
Before this change, running the first command in the AWS CLI doesn't work (that's what the code runs),  and running the second command does work.
```
aws secretsmanager get-secret-value --secret-id elliotgreenlee-dev-civiform-pgadmin-default-username

aws secretsmanager get-secret-value --secret-id elliotgreenlee-dev-cf-pgadmin-default-username
```

dkatzz (approved):

PR #285, opened 2024-01-30
Create terraform variable for secret length

dkatzz:
https://github.com/civiform/civiform/issues/6506

We want to create a terraform variable for the random password length, so certain deployments can update it before they have users.

We'll need to update this before the play framework migration (https://www.playframework.com/documentation/2.9.x/Migration29#Application-Secret-enforces-a-minimum-length), and it will be easier to change the length for the deployments without many users now.

### DIFF
--- a/cloud/aws/templates/aws_oidc/secrets.tf
+++ b/cloud/aws/templates/aws_oidc/secrets.tf
@@ -79,7 +79,7 @@ resource "aws_secretsmanager_secret_version" "postgres_password_secret_version" 
 
 # Create a random generated password to use for app_secret_key.
 resource "random_password" "app_secret_key" {
-  length           = 16
+  length           = var.random_password_length
   special          = true
   override_special = "_%@"
 }
--- a/cloud/aws/templates/aws_oidc/variables.tf
+++ b/cloud/aws/templates/aws_oidc/variables.tf
@@ -472,3 +472,9 @@ variable "monitoring_stack_enabled" {
   description = "If true, Prometheus and Grafana instances are created."
   default     = true
 }
+
+variable "random_password_length" {
+  type        = number
+  description = "Length of the random generated password to use for app_secret_key"
+  default     = 16
+}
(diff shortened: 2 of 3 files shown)


cloud/aws/templates/aws_oidc/variables.tf
```diff
@@ -472,3 +472,9 @@ variable "monitoring_stack_enabled" {
   description = "If true, Prometheus and Grafana instances are created."
   default     = true
 }
+
+variable "random_password_length" {
+  type        = number
+  description = "Length of the random generated password to use for app_secret_key"
+  default     = 16
```
gwendolyngoetz, line 479: The default should be at least 32 or Play won't start once we merge in v2.9. I'd suggest making it 64 so we're set if we ever need to increate the signature algorithm from HS256 to HS512.

cloud/aws/templates/aws_oidc/secrets.tf
```diff
@@ -79,7 +79,7 @@ resource "aws_secretsmanager_secret_version" "postgres_password_secret_version"
 
 # Create a random generated password to use for app_secret_key.
 resource "random_password" "app_secret_key" {
-  length           = 16
+  length           = var.random_password_length
```
gwendolyngoetz, line 82: Is there a way to error if this is smaller than the default?

gwendolyngoetz (changes requested):

cloud/aws/templates/aws_oidc/variables.tf
dkatzz: I just don't want to break any existing deployments with the session stuff, but I think the plan is to change this default to higher once we know the implications and then do the play migration.

cloud/aws/templates/aws_oidc/secrets.tf
dkatzz: Yeah! I can add a validation, but we can do that once we update this default to the higher one.

gwendolyngoetz (approved):

PR #291, opened 2024-03-11
Check in JMeter tests

dkatzz:
### Description

We created JMeter tests for load testing work and should ensure these are checked in to allow others to run tests on their work and in the future run performance tests automatically.

This also adds a new trusted intermediary test.

I'll plan to create documentation to explain how to run these tests.

Tested locally by running the following for each test:

- testing using jmeter: `jmeter  -n -t load-test/applicant_landing_page.jmx` (maybe needs jmeter install)
- testing using bzt: `pip install bzt` `bzt load-test/applicant_landing_page.jmx`

#### General

- [x] Added the correct label

### DIFF
--- a/load-test/applicant_landing_page.jmx
+++ b/load-test/applicant_landing_page.jmx
@@ -1,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jmeterTestPlan version="1.2" properties="5.0" jmeter="5.5">
+  <hashTree>
+    <TestPlan guiclass="TestPlanGui" testclass="TestPlan" testname="Test Plan" enabled="true">
+      <stringProp name="TestPlan.comments"></stringProp>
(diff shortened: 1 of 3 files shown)


gwendolyngoetz:
@dkatzz When you have a chance please update the wiki with instructions on how to run this and how to install jmeter.

gwendolyngoetz (approved):

gwendolyngoetz:
@dkatzz Just realized this is not `civiform/civiform` should this go over there instead?

dkatzz:
> @dkatzz Just realized this is not `civiform/civiform` should this go over there instead?

I was originally thinking we'd use these for load tests in AWS, but maybe worth putting them into the civiform repo so we can run performance tests on different branches, etc. I'll go ahead and close this PR for now and move this over to the other repo if that makes more sense to you?

dkatzz:
Tracking in https://github.com/civiform/civiform/pull/6866 instead

PR #293, opened 2024-03-14
Add rotate_app_secret command

nb1701:
Using bin/run, you can now use the "rotate_app_secret" command. This will replace the value of the app secret key and redeploy CiviForm. When this happens, all sessions will become invalidated, so any guest uses with applications in progress will lose their application, any logged in users or admins will need to log in again.

#### General

- [x] Added the correct label

Run `bin/run` with `rotate_app_secret`, verify in AWS the secret value has changed, verify the staging environment still works.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/6197

### DIFF
--- a/cloud/shared/bin/rotate_app_secret.py
+++ b/cloud/shared/bin/rotate_app_secret.py
@@ -15,9 +15,9 @@ def run(config: ConfigLoader, _params: List[str]):
             f'cloud.{config.get_cloud_provider()}.bin.rotate_app_secret')
         if sys.stdin.isatty():
             answer = input(
-                "WARNING: Rotating the app secret will invalidate all existing sessions. Any guest users with an unsubmitted application in progress will lose their application, and logged in users and admins will need to log in again. Are you sure you want to continue? [Y/n]: "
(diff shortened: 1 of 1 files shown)


avaleske (approved):

cloud/shared/bin/rotate_app_secret.py
```diff
@@ -0,0 +1,27 @@
...
+        'cloud', config.get_cloud_provider(), 'bin', 'rotate_app_secret.py')
+    # TODO(#2741): remove the fork after we remove non-Python scripts
+    if os.path.exists(rotate_file_py):
+        rotate_module = importlib.import_module(
+            f'cloud.{config.get_cloud_provider()}.bin.rotate_app_secret')
+        if sys.stdin.isatty():
+            answer = input(
+                "WARNING: Rotating the app secret will invalidate all existing sessions. Any guest users with an unsubmitted application in progress will lose their application, and logged in users and admins will need to log in again. Are you sure you want to continue? [Y/n]: "
+            )
+            if answer.lower() in ['y', 'yes', '']:
```
rockycodes, line 20: suggestion (non-blocking): Does including `''` mean that if they hit `enter` it will run? If so, I would remove that option so they don't accidentally trigger it by trying to exit the dialogue or something.

rockycodes: That's my preference since I would prefer to accidentally exit the script than accidentally log everyone out.

PR #297, opened 2024-03-29
Setup github codespaces for cloud-deploy-infra repo

dkatzz:
### Description

This change creates a devcontainer for the cloud-deploy-infra repo and adds a dockerfile that installs terraform and the other required dependencies. I tried this and was able to get it working with running both bin commands.

I had to make a small change to `bin/fmt` because it doesn't look like the `-s` flag is recognized but tested and works as expected without it. 

#### General

- [x] Added the correct label

Checkout the branch and click code and then create a codespace and test out commands

### Issue(s) this completes

https://github.com/civiform/civiform/issues/6999

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,20 @@
+FROM mcr.microsoft.com/devcontainers/base:ubuntu
+
+RUN wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | tee /usr/share/keyrings/hashicorp-archive-keyring.gpg
+RUN echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | tee /etc/apt/sources.list.d/hashicorp.list
+
+# Install Terraform
+RUN apt-get update && apt-get install -y \
+    wget \
+    unzip \
+    terraform \
+    direnv \
+    && rm -rf /var/lib/apt/lists/*
+
+# Set up direnv hook
+RUN echo 'eval "$(direnv hook bash)"' >> ~/.bashrc
+RUN echo 'eval "$(direnv hook bash)"' >> ~/.bash_profile
+
+# Install pip and yapf
+RUN apt-get update && apt-get install -y python3-pip
+RUN pip install yapf
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -5,16 +5,13 @@ RUN echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] http
 
 # Install Terraform
 RUN apt-get update && apt-get install -y \
-    wget \
-    unzip \
     terraform \
-    direnv \
-    && rm -rf /var/lib/apt/lists/*
+    python3-pip \
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /var/lib/apt/lists.d/* \
+    && apt-get autoremove \
+    && apt-get clean \
+    && apt-get autoclean
 
-# Set up direnv hook
-RUN echo 'eval "$(direnv hook bash)"' >> ~/.bashrc
-RUN echo 'eval "$(direnv hook bash)"' >> ~/.bash_profile
-
-# Install pip and yapf
-RUN apt-get update && apt-get install -y python3-pip
+# Install yapf
 RUN pip install yapf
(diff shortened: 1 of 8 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | tee /usr/share/keyrings/hashicorp-archive-keyring.gpg
+RUN echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | tee /etc/apt/sources.list.d/hashicorp.list
+
+# Install Terraform
+RUN apt-get update && apt-get install -y \
+    wget \
+    unzip \
+    terraform \
+    direnv \
+    && rm -rf /var/lib/apt/lists/*
```
gwendolyngoetz, line 12:
```suggestion
    && rm -rf /var/lib/apt/lists/* \
    && rm -rf /var/lib/apt/lists.d/* \
    && apt-get autoremove \
    && apt-get clean \
    && apt-get autoclean
```

This will clean out other things from that layer

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM mcr.microsoft.com/devcontainers/base:ubuntu
+
+RUN wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | tee /usr/share/keyrings/hashicorp-archive-keyring.gpg
+RUN echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | tee /etc/apt/sources.list.d/hashicorp.list
+
+# Install Terraform
+RUN apt-get update && apt-get install -y \
+    wget \
+    unzip \
```
gwendolyngoetz, line 9: (non-blocking) Since `wget` is used earlier you probably don't need it here. I suspect unzip may be part of the base ubuntu image as well.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,48 @@
...
+                    "**/.bloop": true,
+                    "**/node_modules": true,
+                    "**/__pycache__": true
+                },
+                "search.exclude": {
+                    "*/target": true,
+                    "**/.bloop": true,
+                    "**/node_modules": true,
+                    "**/__pycache__": true
+                },
```
gwendolyngoetz, line 31: (question) Are there any terraform directories to ignore?

bin/fmt
```diff
@@ -2,23 +2,23 @@
 
 # DOC: Format terraform, python, and shell script files.
 
-if which -s terraform; then
+if which terraform; then
```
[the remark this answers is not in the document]
dkatzz: I was able to get it working on my mac still without the `-s`, but let me know if you think there is anything else worth testing!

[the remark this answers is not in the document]
gwendolyngoetz: Awesome!

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    terraform \
+    direnv \
+    && rm -rf /var/lib/apt/lists/*
+
+# Set up direnv hook
+RUN echo 'eval "$(direnv hook bash)"' >> ~/.bashrc
+RUN echo 'eval "$(direnv hook bash)"' >> ~/.bash_profile
+
+# Install pip and yapf
+RUN apt-get update && apt-get install -y python3-pip
```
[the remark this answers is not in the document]
dkatzz: Good call, updated!

.devcontainer/devcontainer.json
dkatzz: I actually realized I didn't need this at all since we just have the config files in this directory and don't build anything here

.devcontainer/Dockerfile
dkatzz: Yes you're right good catch! Removed

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN apt-get update && apt-get install -y \
+    wget \
+    unzip \
+    terraform \
+    direnv \
+    && rm -rf /var/lib/apt/lists/*
+
+# Set up direnv hook
+RUN echo 'eval "$(direnv hook bash)"' >> ~/.bashrc
+RUN echo 'eval "$(direnv hook bash)"' >> ~/.bash_profile
```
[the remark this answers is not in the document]
dkatzz: I found it in step 8 here https://medium.com/@gyulshenabazz/mastering-terraform-with-visual-studio-code-dev-containers-5294e3f5bfaa used to manage terraform environment variables but I realized we don't do that with a .envrc file, so we actually don't need it.

gwendolyngoetz (approved):

PR #298, opened 2024-04-02
Add tagging action

nb1701:
Adds an action to tag the repo. Partially addresses https://github.com/civiform/civiform/issues/4763

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -1,0 +1,27 @@
+name: Tag Repo
+
+on:
+  workflow_dispatch:
+    inputs:
+      commit_sha:
+        description: 'Commit SHA'
+        required: true
+        default: 'HEAD'
+
+      version:
+        description: 'The version for this release, matching the CiviForm release version, e.g. v1.2.3'
+        required: true
+
+jobs:
+  tag:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v4
+          with:
+            fetch-depth: 0
+        - name: Tag the repo
+          run: |
+            git config user.email "github-actions[bot]@users.noreply.github.com"
+            git config user.name "github-actions[bot]"
+            git tag ${{ github.event.inputs.version }} --message "${{ github.event.inputs.version }}"
+            git push origin ${{ github.event.inputs.version }}


.github/workflows/tag.yaml
```diff
@@ -0,0 +1,27 @@
...
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v4
+          with:
+            fetch-depth: 0
+        - name: Tag the repo
+          run: |
+            git config user.email "github-actions[bot]@users.noreply.github.com"
+            git config user.name "github-actions[bot]"
+            git tag -a ${{ github.event.inputs.release_number }} -m "Release ${{ github.event.inputs.release_number }}"
```
gwendolyngoetz, line 26:
```suggestion
            git tag ${{ github.event.inputs.release_number }} --message="Release ${{ github.event.inputs.release_number }}"
```

I had to look up the `-a` since I hadn't used it. I'd remove it unless you found a reason it needs to be explicitly set. The docs say:

>  If -m <msg> or -F <file> is given and -a, -s, and -u <keyid> are absent, -a is implied.

gwendolyngoetz (changes requested):
When we get to running this on the release call in civiform/civiform do we want to prevent this from manually being run in order to prevent them from getting out of sync?

.github/workflows/tag.yaml
```diff
@@ -0,0 +1,27 @@
...
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v4
+          with:
+            fetch-depth: 0
+        - name: Tag the repo
+          run: |
+            git config user.email "github-actions[bot]@users.noreply.github.com"
+            git config user.name "github-actions[bot]"
+            git tag ${{ github.event.inputs.version }} -m "${{ github.event.inputs.version }}"
```
gwendolyngoetz, line 26:
```suggestion
            git tag ${{ github.event.inputs.version }} --message="${{ github.event.inputs.version }}"
```

Using the full arg name make it easier to understand at a glance.

gwendolyngoetz (changes requested):

gwendolyngoetz:
> > When we get to running this on the release call in civiform/civiform do we want to prevent this from manually being run in order to prevent them from getting out of sync?
> 
> Is that a thing we can do? Might still be nice to have the option to run it manually in case things go sideways, but I don't know. It's pretty easy to delete a bad tag if we need to.

I thought there was a trigger type that let use trigger it, but not in the ui. I might be remembering wrong.

gwendolyngoetz (approved):

PR #310, opened 2024-04-18
Create dumpdb command

nb1701:
### Description

This adds a new command called "dumpdb" and a new Terraform module called dbaccess. The module creates a temporary EC2 host with a public IP, with a security group that allows SSH access only to the host this script is running from, and with PostgreSQL access to the RDS database. The script sets these resources up, along with a temporary keypair, SSHs into the temporary host to install postgresql-client and run pg_dump, SCPs the dump file to the host running this script, then cleans up all the resources.  This also refactors the CIDR state machine into a separate file. While this command does not use the whole state machine, it uses the function to detect the public IP of the current host.

#### General

- [x] Added the correct label

Run `bin/run` and then `dumpdb`.

### Issue(s) this completes

Addresses the first half of https://github.com/civiform/civiform/issues/7232

### DIFF
--- a/cloud/aws/bin/dumpdb.py
+++ b/cloud/aws/bin/dumpdb.py
@@ -4,9 +4,11 @@ import sys
 import tempfile
 import shlex
 import subprocess
+import textwrap
+import urllib.request
+from pathlib import Path
 from time import sleep
 from datetime import datetime
-import urllib.request
 
 from cloud.aws.templates.aws_oidc.bin import resources
 from cloud.aws.templates.aws_oidc.bin.aws_cli import AwsCli
@@ -15,18 +17,26 @@ from cloud.shared.bin.lib.config_loader import ConfigLoader
 from cloud.shared.bin.lib.print import print
 
 
+class Color:
+    RED = '\033[31m'
+    GREEN = '\033[32m'
+    YELLOW = '\033[33m'
+    END = '\033[0m'
+
+
 def run(config: ConfigLoader):
     aws = AwsCli(config)
     print(
-        "\n"
-        "\033[31m!!! WARNING !!!: This command will create a dump of the entire database, including personally identifiable information (PII). Ensure you take the utmost care in handling this data and store it in a secure location.\033[0m\n"
-        "\n"
-        'This process will set up a temporary EC2 host with access to the database, use SSH to run the pg_dump command on that host, then SCP the file to this machine. '
-        'You will need to confirm the application of the Terraform manifest that creates these temporary resources, and then confirm the teardown of these resources. '
-        'If something goes wrong and this process is interrupted before it tears down the resources, you can find them all with the "Module = dbaccess" tag in the AWS console. They should be deleted manually.'
-        "\n\n"
-        'The "ssh" and "ssh-keygen" commands must be available on your machine, typically provided by the openssh-client package. If you do not have these commands, you will need to install them before proceeding.'
-        "\n\n")
+        textwrap.dedent(
+            f"""
+        {Color.RED}!!! WARNING !!!: This command will create a dump of the entire database, including personally identifiable information (PII). Ensure you take the utmost care in handling this data and store it in a secure location.{Color.END}
+
+        This process will set up a temporary EC2 host with access to the database, use SSH to run the pg_dump command on that host, then SCP the file to this machine. You will need to confirm the application of the Terraform manifest that creates these temporary resources, and then confirm the teardown of these resources.
+        
+        If something goes wrong and this process is interrupted before it tears down the resources, you can find them all with the "Module = dbaccess" tag in the AWS console. They should be deleted manually.
+
+        The "ssh" and "ssh-keygen" commands must be available on your machine, typically provided by the openssh-client package. If you do not have these commands, you will need to install them before proceeding.
+    """))
     answer = input('Do you understand the risks and wish to proceed? (y/N): ')
     if answer.lower() != 'y':
         print('Exiting.')
@@ -35,9 +45,8 @@ def run(config: ConfigLoader):
     timestamp = datetime.now().strftime('%Y-%m-%d_%H-%M-%S')
 
     # Current working dir should be the 'checkout' folder, so go one level above.
-    default_file = os.path.join(
-        os.path.dirname(os.getcwd()),
-        f'{config.app_prefix}_civiform_database_{timestamp}.dump')
+    default_filename = f'{config.app_prefix}_civiform_database_{timestamp}.dump'
+    default_file = Path.cwd().parent / default_filename
     dumpfile = input(
         f"Enter the location to save the dump file (default: {default_file}): "
     ) or default_file
@@ -121,11 +130,11 @@ def run(config: ConfigLoader):
             _run_cmd(cmd)
 
             input(
-                '\033[32mDatabase dump complete. Press Enter to tear down the temporary resources.\033[0m'
+                f'{Color.GREEN}Database dump complete. Press Enter to tear down the temporary resources.{Color.END}'
             )
         except:
             input(
-                "\n\033[31mError occurred. See details above. Press Enter to tear down the temporary resources.\033[0m"
+                f"\n{Color.RED}Error occurred. See details above. Press Enter to tear down the temporary resources.{Color.END}"
             )
             raise
         finally:
@@ -145,7 +154,7 @@ def _detect_public_ip() -> str:
             return ip
     except:
         print(
-            'Unable to find the public IP of this machine using checkip.amazonaws.com.'
+            f'{Color.YELLOW}Unable to find the public IP of this machine using checkip.amazonaws.com.{Color.END}'
         )
         return _ask_for_ip()
 
@@ -157,7 +166,9 @@ def _ask_for_ip() -> str:
             ipaddress.IPv4Address(answer)
             return answer
         except ValueError:
-            print('Invalid IP address. Please try again.')
+            print(
+                f'{Color.YELLOW}Invalid IP address. Please try again.{Color.END}'
+            )
 
 
 def _run_cmd(cmd, quiet=False):
@@ -167,10 +178,12 @@ def _run_cmd(cmd, quiet=False):
             break
         except subprocess.CalledProcessError as e:
             if not quiet:
+                print(Color.RED)
                 print('Error running command:')
                 print("Command:", e.cmd)
                 print("Return code:", e.returncode)
                 print("Output:", e.output.decode())
+                print(Color.END)
             raise e
 
 


cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,142 @@
...
+                        sleep(10)
+                    else:
+                        raise e
+
+            print('Installing postgresql-client')
+            cmd = ssh + '"sudo apt-get update && sudo apt-get install -y postgresql-client"'
+            _run_cmd(cmd)
+
+            print('Generating dump file')
+            cmd = ssh + f"PGPASSWORD='{db_pwd}' pg_dump -w -Fc -h '{db_hostname}' -U '{db_user}' -d postgres > civiform_database.dump"
```
gwendolyngoetz, line 87: (blocking?) This has the same issue as the existing script. The PGPASSWORD is plain text if you view processes 😕

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,142 @@
...
+            _run_cmd(cmd)
+
+            print('Downloading dump file to local machine')
+            cmd = f'scp {args} "ubuntu@{ec2_host_ip}:/home/ubuntu/civiform_database.dump" {dumpfile}'
+            _run_cmd(cmd)
+
+            # Not strictly necessary, but in case the host sticks around for some reason.
+            print('Delete dump file on EC2 host')
+            cmd = ssh + 'rm -f civiform_database.dump'
```
gwendolyngoetz, line 96: (praise) Good idea

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,142 @@
...
+            )
+        except:
+            input(
+                "\nError occurred. See details above. Press Enter to tear down the temporary resources."
+            )
+            raise
+        finally:
+            print('Cleaning up resources')
+            os.unsetenv("TF_VAR_dbaccess_cidr_allowlist")
+            os.unsetenv("TF_VAR_dbaccess")
```
gwendolyngoetz, line 110: (question) Wouldn't these have been set in the python process and go out of scope when the app ends?

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,142 @@
...
+        with urllib.request.urlopen("https://checkip.amazonaws.com",
+                                    timeout=3) as response:
+            # response contains a newline
+            return response.read().decode("ascii").strip()
+    except:
+        return ""
+
+
+def _run_cmd(cmd, quiet=False):
+    attempts = 0
```
gwendolyngoetz, line 125:
```suggestion
```
This doesn't appear to be used anywhere.

cloud/aws/modules/dbaccess/ip.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
```
gwendolyngoetz, line 2: (blocking) Add the `DOC` comment on what this script is does

cloud/aws/modules/dbaccess/ip.sh
```diff
@@ -0,0 +1,11 @@
...
+read -r input
+region=$(echo $input | grep -o '"region": *"[^"]*"' | sed 's/"region": *"\([^"]*\)"/\1/')
+# Perform DNS lookups of hosts needed for updating and installing
+# packages on Ubuntu
+security_ubuntu_com=$(dig +short security.ubuntu.com | head -n 1)
+archive_ubuntu_com=$(dig +short archive.ubuntu.com | head -n 1)
+region_ec2_archive_ubuntu_com=$(dig +short $region.ec2.archive.ubuntu.com | head -n 1)
+
+echo "{\"security_ubuntu_com\": \"${security_ubuntu_com}/32\", \"archive_ubuntu_com\": \"${archive_ubuntu_com}/32\", \"region_ec2_archive_ubuntu_com\": \"${region_ec2_archive_ubuntu_com}/32\"}"
```
gwendolyngoetz, line 11: (question, blocking) If any of the `dig` commands return nothing you'll end up with just `/32`. If that's desired please leave a comment on it. If not check the length of the variables and error if any are empty.

cloud/aws/modules/dbaccess/ip.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+read -r input
+region=$(echo $input | grep -o '"region": *"[^"]*"' | sed 's/"region": *"\([^"]*\)"/\1/')
```
gwendolyngoetz, line 4: (suggestion) Add some comments on what region is expected to be. It says ubuntu so I don't want to assume is the AWS region. Maybe even validate it's an allowed value if it's a fixed enum.

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,142 @@
...
+
+    dumpdir = os.path.dirname(dumpfile)
+    if not os.path.isdir(dumpdir):
+        os.makedirs(dumpdir)
+        print(f'Directory created: {dumpdir}')
+
+    # Generate a new key pair for the dbaccess instance. We'll
+    # save this to a temp directory and run all the critical pieces
+    # inside this block so we ensure the key is cleaned up.
+    with tempfile.TemporaryDirectory() as tmpdir:
```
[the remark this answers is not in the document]
gwendolyngoetz: It may depend on the OS, but on Ubuntu (22.04) files I create in /tmp are 664, so anyone can read them. Can you create the file as just 600?

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,142 @@
...
+def run(config: ConfigLoader):
+    aws = AwsCli(config)
+    print(
+        "\n"
+        "!!! WARNING !!!: This command will create a dump of the entire database, including personally identifiable information (PII). Ensure you take the utmost care in handling this data and store it in a secure location.\n"
+        "\n"
+        'This process will set up a temporary EC2 host with access to the database, use SSH to run the pg_dump command on that host, then SCP the file to this machine. '
+        'You will need to confirm the application of the Terraform manifest that creates these temporary resources, and then confirm the teardown of these resources. '
+        'If something goes wrong and this process is interrupted before it tears down the resources, you can find them all with the "Module = dbaccess" tag in the AWS console. They should be deleted manually.'
+        "\n")
```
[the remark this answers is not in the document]
dkatzz: I think it may be worth prompting after this warning to confirm they still want to proceed knowing the risks?

cloud/aws/bin/dumpdb.py
dkatzz: The difference with pgadmin is that its a temporary password created for the session, but this would be the secret value, right? From what I remember the password I get from pgadmin is different each time

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,142 @@
...
+    # Current working dir should be the 'checkout' folder, so go one level above.
+    default_file = os.path.join(
+        os.path.dirname(os.getcwd()), 'civiform_database.dump')
```
dkatzz, line 28: **suggestion:**  I think its worth including the var.app_prefix in this filename. I can imagine if someone has multiple instances they may get confused between different dumps. You can also consider including the date.

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,142 @@
...
+
+            print('Installing postgresql-client')
+            cmd = ssh + '"sudo apt-get update && sudo apt-get install -y postgresql-client"'
+            _run_cmd(cmd)
+
+            print('Generating dump file')
+            cmd = ssh + f"PGPASSWORD='{db_pwd}' pg_dump -w -Fc -h '{db_hostname}' -U '{db_user}' -d postgres > civiform_database.dump"
+            _run_cmd(cmd)
+
+            print('Downloading dump file to local machine')
```
dkatzz, line 90: **suggestion:** I assume you'll have to run this command from the terminal on your computer (or a vm or something) and can't run it on cloudshell or something because then the file would be kind of stuck in that cloudshell instance related to that particular AWS deployment? If that's the case, I think it could be worth including a separate prompt after the warning ensuring that the user is running this from the right place. We may also want to check where the deployers of cf today run their deployment from to confirm this would work for them.

cloud/aws/modules/dbaccess/main.tf
```diff
@@ -0,0 +1,91 @@
+# Needed so the EC2 host can update and install packages
+# AWS only allows specifying by IP, not hostname, and Terraform
+# doesn't have a way to look up the IP of a hostname. This is
+# kind of brittle, but seems like the safest way so we don't
+# have to allow access to any IP.
+data "external" "external_ips" {
```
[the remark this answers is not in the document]
gwendolyngoetz: Let's keep it for now and if it becomes a problem relax the security or figure out another approach later.

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,169 @@
...
+        "\n\n")
+    answer = input('Do you understand the risks and wish to proceed? (y/N): ')
+    if answer.lower() != 'y':
+        print('Exiting.')
+        return
+
+    timestamp = datetime.now().strftime('%Y-%m-%d_%H-%M-%S')
+
+    # Current working dir should be the 'checkout' folder, so go one level above.
+    default_file = os.path.join(
```
dkatzz, line 38: **suggestion:** This is being set, but it doesn't seem like we're actually using it when storing the file? I think we should use this naming on both the EC2 and local filesystem

cloud/aws/bin/dumpdb.py
dkatzz: Makes sense - probably worth understanding where each entity runs their deployment from so we can make sure we support all the methods.

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,180 @@
...
+from cloud.shared.bin.lib import terraform
+from cloud.shared.bin.lib.config_loader import ConfigLoader
+from cloud.shared.bin.lib.print import print
+
+
+def run(config: ConfigLoader):
+    aws = AwsCli(config)
+    print(
+        "\n"
+        "\033[31m!!! WARNING !!!: This command will create a dump of the entire database, including personally identifiable information (PII). Ensure you take the utmost care in handling this data and store it in a secure location.\033[0m\n"
```
yegle, line 22: Nit: the usual recommendation of output a block of text is to use triple-quotes and `textwrap.dedent`:

```
print(textwrap.dedent("""
    Note this is a long
    multi-line
    message
"""))
```

cloud/aws/bin/dumpdb.py
```diff
@@ -0,0 +1,180 @@
...
+        "\n\n")
+    answer = input('Do you understand the risks and wish to proceed? (y/N): ')
+    if answer.lower() != 'y':
+        print('Exiting.')
+        return
+
+    timestamp = datetime.now().strftime('%Y-%m-%d_%H-%M-%S')
+
+    # Current working dir should be the 'checkout' folder, so go one level above.
+    default_file = os.path.join(
```
yegle, line 38: Nit: `pathlib` is more readable.

```python
from pathlib import Path

default_file = Path('.') / f'{config.app_prefix}_civiform_database_{timestamp}.dump'
```

dkatzz (approved):

gwendolyngoetz (approved):

PR #311, opened 2024-04-23
Add restoredb command

nb1701:
This adds a restoredb command, which works very similarly to the dumpdb command.  It sets up an EC2 host with access to the database, SCPs the dump file to it, and runs pg_restore.

This also modifies dbaccess a bit. Instead of finding IPs with dig, we now just allow 0.0.0.0/0 egress on 80 and 443. We need to install version 16 of the postgresql client in order for this to work once we upgrade to PG16, and this involves reaching out more places, and reaching out with curl.  Curl isn't as resilient to trying different available IPs for a domain like apt is, so we'd have to allow every IP for apt.postgresql.org.  Since this part of the code was brittle anyway, and allowing egress (but not ingress!) to whatever is probably relatively okay, this has all been simplified.

This also fixes a bug when not running dumpdb/restoredb, in that the dbaccess_public_key parameter needs some default so it doesn't get in the way when running bin/setup or bin/deploy.

Addresses the rest of https://github.com/civiform/civiform/issues/7232


cloud/aws/bin/restoredb.py
```diff
@@ -0,0 +1,197 @@
...
+from datetime import datetime
+
+from cloud.aws.templates.aws_oidc.bin import resources
+from cloud.aws.templates.aws_oidc.bin.aws_cli import AwsCli
+from cloud.shared.bin.lib import terraform
+from cloud.shared.bin.lib.config_loader import ConfigLoader
+from cloud.shared.bin.lib.print import print
+
+
+class Color:
```
dkatzz, line 20: **suggestion:** Since we're using this in other places, we could consider pulling it out into a separate file?

PR #318, opened 2024-05-16
Throw error if ADMIN_WRITEABLE vars in config

rockycodes:
### Description

Variables that are marked as `ADMIN_WRITEABLE` in [env-var-docs.json](https://github.com/civiform/civiform/blob/main/server/conf/env-var-docs.json) are editable in the admin settings config and should not be set in the deploy configs. This change adds an error if an `ADMIN_WRITEABLE` is found in the deploy config. It also allows the ability to override this by setting `ALLOW_ADMIN_WRITEABLE=true` in the deploy config.

Notes:
- Need to check all of the config files we manage to see if we need to add `ALLOW_ADMIN_WRITEABLE=true` (I know we need to do this for staging)
- Need to update documentation
- Need to provide guidance to CEs that they might see this error if they have `ADMIN_WRITEABLE` vars in their deploy config

#### General

- [x] Added the correct label
- [x] Extended the README / documentation, if necessary https://github.com/civiform/docs/pull/459

1. Add an admin writeable variable to your deploy config (Eg. `WHITELABEL_CIVIC_ENTITY_FULL_NAME`)
2. Run `bin/setup` (or `bin/deploy` if you already have a running deployment)
3. See error
4. Add `ALLOW_ADMIN_WRITEABLE=true` to deploy config
5. Run `bin/setup` (or `bin/deploy`) again.
6. Deploy should continue as normal with no error

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/6663

### DIFF
--- a/cloud/shared/bin/lib/config_loader.py
+++ b/cloud/shared/bin/lib/config_loader.py
@@ -91,6 +91,7 @@ class ConfigLoader:
             answer = input(msg)
             if answer not in ['y', 'Y', 'yes']:
                 exit(1)
+
         self._export_env_variables(config_fields)
         return config_fields
 
--- a/cloud/shared/bin/lib/config_loader.py
+++ b/cloud/shared/bin/lib/config_loader.py
@@ -87,9 +87,9 @@ class ConfigLoader:
 
                 Would you like to continue with the deploy? [y/N] > 
                 """)
-        answer = input(msg)
-        if answer not in ['y', 'Y', 'yes']:
-            exit(1)
+            answer = input(msg)
+            if answer not in ['y', 'Y', 'yes']:
+                exit(1)
         self._export_env_variables(config_fields)
         return config_fields
 
--- a/cloud/shared/bin/lib/config_loader.py
+++ b/cloud/shared/bin/lib/config_loader.py
@@ -347,7 +347,8 @@ class ConfigLoader:
                     int(config_value)
                 except ValueError as e:
                     validation_errors.append(
-                        f'{Color.RED}{name} is required to be an integer: {e}{Color.END}')
+                        f'{Color.RED}{name} is required to be an integer: {e}{Color.END}'
+                    )
                     continue
 
             if variable.type == "bool":
--- a/cloud/shared/bin/lib/config_loader.py
+++ b/cloud/shared/bin/lib/config_loader.py
@@ -70,6 +70,26 @@ class ConfigLoader:
         """
         config_parser = ConfigParser()
         config_fields = config_parser.parse_config(config_file)
+
+        if config_fields.get(
+                "ALLOW_ADMIN_WRITEABLE") and not os.getenv('SKIP_USER_INPUT'):
+            msg = inspect.cleandoc(
+                """
+                ###########################################################################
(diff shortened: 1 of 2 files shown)


dkatzz (approved):
Overall LGTM! In which cases do we expect CEs to set ADMIN_WRITEABLE=true? My only concern is that a CE would set that to get around the error then forget about it, so I wonder if its worth including some sort of warning message if that's set to true. Or maybe removing the ADMIN_WRITEABLE option all together?

dkatzz:
Would you mind changing the PR title to be more descriptive?

rockycodes:
> Overall LGTM! In which cases do we expect CEs to set ADMIN_WRITEABLE=true? My only concern is that a CE would set that to get around the error then forget about it, so I wonder if its worth including some sort of warning message if that's set to true. Or maybe removing the ADMIN_WRITEABLE option all together?

Good idea, I can add a warning message about that. The two cases that I can think of where we would want to allow this are:
1. In our staging environment where we set some custom `ADMIN_WRITEABLE` vars on deployment
2. If a CE wants to deploy for the first time with some custom values for `ADMIN_WRITEABLE` vars

And thanks for noticing my PR title! I'll update that.

cloud/shared/bin/lib/config_loader.py
```diff
@@ -70,6 +71,27 @@ def _load_config_fields(self, config_file: str):
         """
         config_parser = ConfigParser()
         config_fields = config_parser.parse_config(config_file)
+
+        if config_fields.get(
+                "ALLOW_ADMIN_WRITEABLE") and not os.getenv('SKIP_USER_INPUT'):
+            msg = inspect.cleandoc(
+                """
+                ###########################################################################
```
dkatzz, line 79: **nitpick:**  Thanks for adding this! You could also consider using the colors like [here](https://github.com/civiform/cloud-deploy-infra/blob/c6ad29d40036edca377d7d9caf84c602a2ccd8d1/cloud/aws/bin/deploy.py#L68).

cloud/shared/bin/lib/config_loader.py
```diff
@@ -70,6 +71,27 @@ def _load_config_fields(self, config_file: str):
...
+                be set to true if this is the first time you are deploying and you want the custom
+                values for ADMIN_WRITEABLE to be applied. If this is not the case, 
+                please remove ALLOW_ADMIN_WRITEABLE and any ADMIN_WRITEABLE variables from 
+                your deploy config and update ADMIN_WRITEABLE variables via the admin settings
+                panel.
+
+                Would you like to continue with the deploy? [y/N] > 
+                """)
+            answer = input(msg)
+            if answer not in ['y', 'Y', 'yes']:
```
dkatzz, line 92: **suggestion:** You could consider updating to `answer.lower() not in ['y', 'yes']`, so we accept any variation of these

dkatzz (approved):

cloud/shared/bin/lib/config_loader.py
rockycodes: Ooooo I love that idea.

PR #321, opened 2024-05-20
Fix upgrade paths for different postgres versions

nb1701:
It would seem that RDS will let 12.17 ONLY upgrade to 16.1, and 12.18 ONLY upgrade to 16.2, not 16.1. This means we have to maintain a table of valid upgrade paths here until everyone has made it onto 16, at which point RDS will upgrade minor versions automatically.

This changes the deploy logic, making it a bit less generic in order to handle this additional complexity of specific major.minor valid upgrade paths.

#### General

- [x] Added the correct label

You can just deploy, say, 1.62.0 which is on 12, which should set you up with 12.18, and verify this tool upgrades you to 16.2.  The failure we noted before was someone on 12.17, so to do this, you would need to edit main.tf to deploy 12.17 specifically when you're on something like 1.62.0, then perform the upgrade.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/7516

### DIFF
--- a/cloud/aws/bin/deploy.py
+++ b/cloud/aws/bin/deploy.py
@@ -97,7 +97,7 @@ def _check_for_postgres_upgrade(config: ConfigLoader, aws: AwsCli):
                 exit(1)
             if current_version not in upgrade_path.keys():
                 print(
-                    f'{Color.YELLOW}This version of the deployment tool does not have information about if {current_version} is sufficiently new enough to upgrade to the latest minor version of version PostgreSQL {major_to_apply}. Check https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html and verify this is a valid upgrade path.{Color.END}'
+                    f'{Color.YELLOW}This version of the CiviForm deployment tool has no information about your current PostgreSQL version of {current_version} and thus cannot choose the correct version to upgrade to. If you proceed, we will attempt to upgrade to the latest {major_to_apply}.x version, but this may not succeed. Check https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html to verify that this is a valid upgrade path.{Color.END}'
                 )
                 if os.getenv('SKIP_USER_INPUT'):
                     print('Proceeding since SKIP_USER_INPUT is set.')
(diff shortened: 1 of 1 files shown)


cloud/aws/bin/deploy.py
```diff
@@ -45,43 +45,59 @@ def _check_application_secret_length(config: ConfigLoader, aws: AwsCli):
...
+                    f'{Color.RED}In order to upgrade to PostgreSQL {major_to_apply}, you must first upgrade to PostgreSQL 12.17 or a later PostgreSQL 12 minor version. You will need to perform this upgrade in the AWS RDS console before proceeding.{Color.END}'
                 )
-            if current_major not in pg_upgrade_table[to_apply]:
+                exit(1)
+            if current_version not in upgrade_path.keys():
                 print(
-                    f'{Color.YELLOW}This version of the deployment tool does not have information about if {current_major}.{current_minor} is sufficiently new enough to upgrade to version {to_apply}. Check https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html and verify this is a valid upgrade path.{Color.END}'
+                    f'{Color.YELLOW}This version of the deployment tool does not have information about if {current_version} is sufficiently new enough to upgrade to the latest minor version of version PostgreSQL {major_to_apply}. Check https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html and verify this is a valid upgrade path.{Color.END}'
```
gwendolyngoetz, line 100:
```suggestion
                    f'{Color.YELLOW}The CiviForm deployment tool does not yet know if {current_version} is sufficiently new enough to upgrade to the latest minor release of PostgreSQL version {major_to_apply}. Check https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html and verify this is a valid upgrade path.{Color.END}'
```

This was hard to read with the multiple ambiguous use of the work version. Does this still mean what you wanted?

cloud/aws/bin/deploy.py
```diff
@@ -91,21 +107,27 @@ def _check_for_postgres_upgrade(config: ConfigLoader, aws: AwsCli):
...
+
             # If a user sets ALLOW_POSTGRESQL_UPGRADE in their config file, config.get_config_var will pick it up.
             # If they've set it as an environment variable, we need to detect that and then add it to the config
             # object ourselves so that it is picked up with the manifest is compiled.
-            if config.get_config_var("ALLOW_POSTGRESQL_UPGRADE") != "true":
+            if config.get_config_var('ALLOW_POSTGRESQL_UPGRADE') != 'true':
                 if os.getenv('SKIP_USER_INPUT'):
-                    print('Proceeding since SKIP_USER_INPUT is set.')
+                    print(
+                        'Proceeding with upgrade to since SKIP_USER_INPUT is set.'
```
gwendolyngoetz, line 121:
```suggestion
                        'Proceeding with upgrade since SKIP_USER_INPUT is set.'
```

gwendolyngoetz (changes requested):

cloud/aws/bin/deploy.py
gwendolyngoetz: I like your new copy much more. Ship it!

gwendolyngoetz (approved):

PR #330, opened 2024-05-31
Revert "Revert "Provide a knob to optionally skip creating the VPC re…

yegle:
…lated resources" (#328)"

This reverts commit ed78f28d7d8fa10485b4a06de0274d524b822ebe.

This also fixed the breakage originally introduced in f1e35656ec2d148d23ae087aca25068ffe6be5c6

### DIFF
--- a/cloud/aws/templates/aws_oidc/app.tf
+++ b/cloud/aws/templates/aws_oidc/app.tf
@@ -317,13 +317,13 @@ module "ecs_fargate_service" {
   desired_count             = var.fargate_desired_task_count
   default_certificate_arn   = var.ssl_certificate_arn
   ssl_policy                = "ELBSecurityPolicy-FS-1-2-Res-2020-10"
-  vpc_id                    = module.vpc.vpc_id
+  vpc_id                    = local.vpc_id
   task_definition_arn       = var.monitoring_stack_enabled ? aws_ecs_task_definition.civiform_with_monitoring.arn : aws_ecs_task_definition.civiform_only.arn
   container_name            = "${var.app_prefix}-civiform"
   ecs_cluster_name          = module.ecs_cluster.aws_ecs_cluster_cluster_name
   ecs_cluster_arn           = module.ecs_cluster.aws_ecs_cluster_cluster_arn
-  private_subnets           = module.vpc.private_subnets
-  public_subnets            = module.vpc.public_subnets
+  private_subnets           = local.vpc_private_subnets
+  public_subnets            = local.vpc_public_subnets
   max_cpu_threshold         = var.ecs_max_cpu_threshold
   min_cpu_threshold         = var.ecs_min_cpu_threshold
   max_cpu_evaluation_period = var.ecs_max_cpu_evaluation_period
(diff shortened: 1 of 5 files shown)


dkatzz:
Thanks for fixing this, @yegle! Would you mind updating the title to remove the "revert" parts. And then in the description could you include a bit more information about what changed and also include the github issue it relates to (similar to what was included in the original [PR](https://github.com/civiform/cloud-deploy-infra/pull/327), but with some information about what changed between that PR and this one to fix the issue)? And could you also include in the description how you tested it?

yegle:
PR description updated, PTAL.

dkatzz:
Could we try testing with `vars.external_vpc` set for a `bin/setup`? In that case there won't be existing resources. You can first [nuke](https://github.com/civiform/civiform/wiki/Deploy-System#setup-aws-nuke) the resources in the environment or use a different app-prefix.

dkatzz (approved):

yegle:
With the `moved` block added in the last commit, I did the following test:

**Testing w/o external vpc configs:**

1. Deploy to my dev project using the `main` branch
2. Switch to the `fix-vpc-knob` branch (this PR), run `terraform plan`

Output:
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the
following symbols:
  ~ update in-place
-/+ destroy and then create replacement
...
Plan: 2 to add, 1 to change, 2 to destroy.
```

(`resource.aws_ecs_task_definition.civiform_only` and `resource.aws_ecs_task_definition.civiform_with_monitoring` were force replaced due to the image version change)

I did not test w/ external vpc configs. We will need to work with the CE on this, since if the CE partially sets up the VPC using the Terraform config and then sets the `external_vpc` variable, it will destroy all vpc related resources, even if you want to reuse the partially created resource.

When the CE decides to use this knob, we should work closely with them, run `terraform state rm` for every resource that is proposed to destroy and then switch to the external managed VPC resources.

SGTY?

dkatzz:
> With the `moved` block added in the last commit, I did the following test:
> 
> **Testing w/o external vpc configs:**
> 
> 1. Deploy to my dev project using the `main` branch
> 2. Switch to the `fix-vpc-knob` branch (this PR), run `terraform plan`
> 
> Output:
> 
> ```
> Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the
> following symbols:
>   ~ update in-place
> -/+ destroy and then create replacement
> ...
> Plan: 2 to add, 1 to change, 2 to destroy.
> ```
> 
> (`resource.aws_ecs_task_definition.civiform_only` and `resource.aws_ecs_task_definition.civiform_with_monitoring` were force replaced due to the image version change)
> 
> I did not test w/ external vpc configs. We will need to work with the CE on this, since if the CE partially sets up the VPC using the Terraform config and then sets the `external_vpc` variable, it will destroy all vpc related resources, even if you want to reuse the partially created resource.
> 
> When the CE decides to use this knob, we should work closely with them, run `terraform state rm` for every resource that is proposed to destroy and then switch to the external managed VPC resources.
> 
> SGTY?

We'll actually be using a different AWS account for them for this deployment than their previous deployment, so we shouldn't have to do the `terraform state rm`. I think we should be able to just add the new VPC information before the deployment and then see how that goes.

PR #335, opened 2024-06-11
Allow setting the external vpc related variables from config script.

yegle:
### Description

This PR allows the external vpc related configs to be set from the `civiform_config.sh` script.

#### General

- [ ] Added the correct label

### Issue(s) this completes


dkatzz:
> Totally fine as is, but if we're also going to be allowing the private and public subnets to be an array, it might make sense to do that here as well.

Agreed! I'll go ahead and merge this and I can work on the change to update the private and public subnet variables to lists.

PR #341, opened 2024-06-17
Look up AMI to use for dumpdb/restoredb

nb1701:
The AMI used as the default for the dbaccess image was only present in us-east-1, and each region has a different AMI for the same image. This looks up the latest AMI for Ubuntu 24.04 in whatever region you are running in and uses that rather than a hardcoded value.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/7714


rockycodes (commented):
lgtm! smart change.

dkatzz (approved):
Thanks for updating this!

PR #342, opened 2024-07-02
Default to not printing pgadmin/database credentials for pgadmin

nb1701:
Because particularly the database password is sensitive, this now defaults to not printing the pgadmin and database credentials and giving the name of the secret instead so the user can look it up however they wish. But the user may also allow the script to print the credential if they are not worried about the security of printing it to the terminal.

This also refactors the Color lib to make it less cumbersome to use.

Tested with an existing deployment to verify it all works correctly, including the new Color refactor.

### Issue(s) this completes

https://github.com/civiform/civiform/issues/7956


dkatzz (approved):
LGTM! And thanks for doing that color refactor - much cleaner!

PR #366, opened 2024-08-13
Update Azure terraform resources to work with the deployment script

jcma-google:
### Description

- Updated postgresql_server to a postgresql_flexible_server along with necessary attendant changes. the flexible_server is required to support the version of postgres civiform uses.   
   - increasing size of postgres database from 5GB to 32GB. 32 is the minimum for flexible dbs and considering it's much higher than the current size of 5GB it's probably fine as a default.
- Removed canary slots (couldn't get it to work and didn't seem particularly useful when considering  how we recommend civic entities to use one RG per deployment type)
- Updated azurerm_app_service  to azurerm_linux_web_app because azurerm_app_service is being deprecated 
- Misc resource fixes to get them working 
- Updated variables to reference the correct values 
- Updated DB_JDBC_STRING to reference the correct connecting string 
- Remove server application variables from the terraform script
- Fixes authentication by adding in the missing necessary variables

#### General

- [x] Added the correct label
- [x] Extended the README / documentation, if necessary


cloud/azure/modules/app/variables.tf
```diff
@@ -59,13 +52,9 @@ variable "bastion_address_prefixes" {
 }
 
 variable "app_sku" {
-  type        = map(string)
+  type        = string
   description = "SKU tier/size/capacity information"
-  default = {
```
dkatzz, line 64: **question:** Why don't we need the tier and capacity anymore?

cloud/azure/modules/app/variables.tf
```diff
@@ -59,13 +52,9 @@ variable "bastion_address_prefixes" {
 }
 
 variable "app_sku" {
-  type        = map(string)
+  type        = string
   description = "SKU tier/size/capacity information"
```
dkatzz, line 56: **suggestion:** We should update this description

cloud/azure/templates/azure_saml_ses/bin/setup.py
```diff
@@ -158,4 +161,4 @@ def _setup_ses(self):
                 "cloud/azure/bin/ses-to-keyvault", "-v", self.key_vault_name,
                 "-u", aws_username
             ],
-            check=True)
+            check=False)
```
dkatzz, line 164: **question:** What does this change do?

cloud/azure/templates/azure_saml_ses/main.tf
```diff
@@ -22,7 +22,8 @@ module "app" {
   postgres_admin_login = var.postgres_admin_login
 
   # note that we must use GP tier
-  postgres_sku_name = "GP_Gen5_2"
+  postgres_sku_name   = "GP_Standard_D2s_v3"
+  postgres_storage_mb = "32768"
```
dkatzz, line 26: **question:** Is this size a recommendation from somewhere? If so it would be helpful to add that as a comment.

cloud/azure/templates/azure_saml_ses/main.tf
```diff
@@ -37,16 +38,7 @@ module "app" {
...
-  civiform_time_zone_id = var.civiform_time_zone_id
-
-  civic_entity_small_logo_url = var.civic_entity_small_logo_url
-
-  adfs_admin_group = var.adfs_admin_group
-
-  login_radius_api_key       = var.login_radius_api_key
-  login_radius_metadata_uri  = var.login_radius_metadata_uri
-  login_radius_saml_app_name = var.login_radius_saml_app_name
-  saml_keystore_filename     = module.saml_keystore.filename
```
dkatzz, line 49: **question:** Why do we want to remove these?

cloud/azure/templates/azure_saml_ses/main.tf
```diff
@@ -77,13 +67,13 @@ module "saml_keystore" {
...
-  for_each = toset([
-    var.sender_email_address,
-    var.staging_applicant_notification_mailing_list,
-    var.staging_ti_notification_mailing_list,
-    var.staging_program_admin_notification_mailing_list
-  ])
-  source               = "../../../aws/modules/ses"
-  sender_email_address = each.key
-}
+# module "email_service" {
```
dkatzz, line 70: **suggestion:** Could you include a comment that we're removing this because of the reliance on AWS, and it will be replaced with an alternative later? I'll create a bug too we can link here

cloud/shared/bin/lib/terraform.py
```diff
@@ -144,7 +144,7 @@ def perform_apply(
     # Enable compact-warnings as we have a bunch of
     # "value of undeclared variables" warnings as some variables used in one
     # deployment (e.g. aws) but not the other.
-    terraform_apply_cmd = f'terraform -chdir={terraform_template_dir} apply -input=false -var-file={tf_vars_filename} -compact-warnings'
+    terraform_apply_cmd = f'terraform -chdir={terraform_template_dir} apply -input=false -var-file={tf_vars_filename} -compact-warnings -auto-approve'
```
dkatzz, line 147: **question:** Would this cause the changes to be applied without confirming? We can add that as a variable, but we should make the default that they have to approve any changes.

dkatzz (commented):
Thank you SO much again for all your work on this!! Most of my comments are clarifying questions, etc. I think it would be helpful to add a bit more context into the description - i.e. that we're moving to a more recommended postgres server and the change from azurerm_app_service is because that was deprecated. I think ideally we'd split this up (like doing the version changes, etc. in another PR and splitting some of the big changes from each other, like the removal of canary), but that seems like a lot of work, so I think its fine to keep it as is.

cloud/azure/bin/lib/key_vault.sh
```diff
@@ -16,8 +16,7 @@ function key_vault::create_vault() {
   az keyvault create \
     --name "${3}" \
     --resource-group "${1}" \
-    --location "${2}" \
-    --enable-rbac-authorization
```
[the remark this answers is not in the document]
jcma-google: I don't remember now but I think it was because it was throwing an error and I removed it as a test. But I just added it back and it's working now so maybe it was a transient issue.

cloud/azure/modules/app/locals.tf
```diff
@@ -2,8 +2,8 @@ locals {
   # The hard-coded zero indexes here are necessary to access the fqdn from the record set associated with it
   # because the private_dns_zone_configs and record_sets blocks expose lists, even if we only have one dns zone
   # and record set configured.
-  postgres_private_link = azurerm_private_endpoint.endpoint.private_dns_zone_configs[0].record_sets[0].fqdn
-  generated_hostname    = "${var.application_name}-${random_pet.server.id}.azurewebsites.net"
+  # postgres_private_link = azurerm_private_endpoint.endpoint.private_dns_zone_configs[0].record_sets[0].fqdn
```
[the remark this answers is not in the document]
jcma-google: Added a comment as to why I commented it out! 

Reiterating here. I'm planning on coming back and setting up private link so leaving it commented it out for now. Right now the postgres server is password protected, on its own subnet, and firewalled. This feels sufficient for staging purposes.

jeffpw-goog:
> Thank you SO much again for all your work on this!! Most of my comments are clarifying questions, etc. I think it would be helpful to add a bit more context into the description - i.e. that we're moving to a more recommended postgres server and the change from azurerm_app_service is because that was deprecated. I think ideally we'd split this up (like doing the version changes, etc. in another PR and splitting some of the big changes from each other, like the removal of canary), but that seems like a lot of work, so I think its fine to keep it as is.

Big +1 to small CLs :/

cloud/azure/modules/app/outputs.tf
```diff
@@ -1,15 +1,15 @@
 output "app_service_default_hostname" {
-  value = "https://${azurerm_app_service.civiform_app.default_site_hostname}"
-}
-
-output "app_service_canary_hostname" {
-  value = "https://${azurerm_app_service_slot.canary.default_site_hostname}"
+  value = "https://${azurerm_linux_web_app.civiform_app.name}.azurewebsites.net"
```
jeffpw-goog, line 2: **question:** This looks like it'll need to be changed back to allow general URLs - what is the purpose of including the change here?

cloud/azure/templates/azure_saml_ses/main.tf
jeffpw-goog: Related question: do we have any published recommendations on machine size for CiviForm? Is this equivalent to what we use for AWS?

cloud/azure/modules/app/locals.tf
```diff
@@ -32,23 +31,27 @@ locals {
...
 
     # The values below are all defaulted to null. If SAML authentication is used, the values can be pulled from the
     # saml_keystore module
     LOGIN_RADIUS_KEYSTORE_NAME    = (var.saml_keystore_filename != null ? "/saml/${var.saml_keystore_filename}" : "")
     LOGIN_RADIUS_KEYSTORE_PASS    = var.saml_keystore_password
     LOGIN_RADIUS_PRIVATE_KEY_PASS = var.saml_private_key_password
 
-    CIVIFORM_API_SECRET_SALT = data.azurerm_key_vault_secret.api_secret_salt_key.value
+    CIVIFORM_API_SECRET_SALT           = data.azurerm_key_vault_secret.api_secret_salt_key.value
+    DATABASE_APPLY_DESTRUCTIVE_CHANGES = true
```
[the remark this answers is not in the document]
jcma-google: I think this can be removed entirely? Because this variable is inside of this expression 

`app_settings` = merge({
}), var.civiform_server_environmental_variables)`

cloud/azure/templates/azure_saml_ses/bin/setup.py
jcma-google: it basically allows the AWS keyvault portion to fail silently. We don't need the AWS portion if we're not going to use its email service. I would remove _setup_ses during the process but I don't feel super confident in removing it since I'm not sure if something is depending on the keyvault being present. 

The workaround I have right now for the AWS issue is during setup after the keyvault is created going in and placing fake values for the AWS keys.

jcma-google: it basically allows the AWS keyvault portion to fail silently. We don't need the AWS portion if we're not going to use its email service. I would remove _setup_ses during the process but I don't feel super confident in removing it since I'm not sure if something is depending on the keyvault being present. 

The workaround I have right now for the AWS issue is during setup after the keyvault is created going in and placing fake values for the AWS keys.

cloud/azure/bin/lib/key_vault.sh
```diff
@@ -124,5 +124,5 @@ function key_vault::has_secret() {
     --query value \
     -o tsv 2>&1 >/dev/null)"
 
-  echo "${SECRET_RESULT}" | grep -q -v "SecretNotFound"
+  echo ${SECRET_RESULT}
```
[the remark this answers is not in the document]
jcma-google: The old expression doesn't work unfortunately. It always returns true regardless of whether or not "SecretNotFound" is in the string. I don't know enough bash to understand why it's broken. I decided to fix it by writing an expression that I knew would always work because it's a direct check against the returned string.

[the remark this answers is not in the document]
jeffpw-goog: Hmm the old expression works in my shell - I'd be happy to help you debug it! I have a lot of experience with bash :| 

> $ echo "bloo blah" | grep -q -v "SecretNotFound"; echo $?
> 0
> $ echo "bloo blah SecretNotFound" | grep -q -v "SecretNotFound"; echo $?
> 1

If you prefer the new logic can you move it into this function so that it still returns true/false and isn't repeated at the callsite? Either way happy to look at it with you!

cloud/azure/modules/app/outputs.tf
jcma-google: `default_site_hostname` is no longer an attribute in the newer version of the resource `azurerm_linux_web_app` (as compared to the deprecated resource `azurerm_app_service`) but I just realized `default_hostname`.

cloud/azure/templates/azure_saml_ses/main.tf
jcma-google: This is from https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-compute where the minimum storage size for general purpose flexible servers is 32GB. Since this is so much higher than the previous storage amount (5GB) I imagined it would suffice. 

but that does beg the question of what is the current minimum capacity of AWS storage?

cloud/shared/bin/lib/health.sh
```diff
@@ -18,7 +18,7 @@ function health::wait_for_success() {
   echo "Polling ${1} for successful response. This may take a few minutes"
   local health_status="$(health::get_status "${1}")"
 
-  until [[ "${health_status}" -eq "200" ]]; do
+  until [[ "${health_status}" -eq "200" ]] || [[ "${health_status}" -eq "303" ]]; do
```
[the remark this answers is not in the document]
jcma-google: So ${AZURE_PRIMARY_URL}/programs (and `${AZURE_PRIMARY_URL}`) when the app is up and running returns 303 because the initial http response is a redirect to /callback?client_name=GuestClient 

I could curl /callback?client_name=GuestClient directly which would return a 302 but it doesn't seem a whole lot better. 

I'm not quite sure why the initial requset is always a redirect code but I've found that it's a reliable indicator that the app service is up and working, which is what I think this health check is supposed to cover.

cloud/azure/templates/azure_saml_ses/main.tf
jcma-google: Including these variables throw an error, removing them doesn't seem to harm the deployment so far, but I'm also unsure of the importance of these settings. I'll comment them out for now so the setup script runs and come back and fix it as a bug in a separate PR.

dkatzz: Ah these are application variables. We can remove them since they shouldn't need to be set in terraform, but we should test that changing them in the config.sh file then updates them in the settings page:
Screenshot 2024-08-19 at 11 58 27 AM

Can we also add in the description that we removed server application variables from the terraform script?

cloud/azure/modules/app/main.tf
```diff
@@ -55,107 +55,38 @@ resource "azurerm_subnet" "server_subnet" {
...
+  service_plan_id     = azurerm_service_plan.plan.id
+  app_settings        = local.app_settings
 
   site_config {
-    linux_fx_version       = "DOCKER|civiform/civiform:${var.image_tag}"
-    always_on              = true
-    vnet_route_all_enabled = true
-  }
-
-  # We will only mount this storage container if SAML authentication is being used
```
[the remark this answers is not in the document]
jcma-google: I think the diff is confusing because I deleted the canary app service entirely, which included the "dynamic storage account" line. You can see that below on line 157 we do preserve the dynamic storage account but it's muddled between what's kept and what's deleted between the two app service definitions.

cloud/shared/bin/lib/health.sh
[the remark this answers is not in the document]
jeffpw-goog: Thanks for explaining - it looks like if you don't have a session it always redirects you to the callback to establish a guest one (I can repro in an incognito window). My concern with the change is that it's a shared script and the name "wait_for_success" sounds like it would wait for 200 rather than 303. Looks like it's not used anywhere else, though, and the 303 behavior is not Azure-specific, so I guess it's okay. Maybe add a comment as to why 303 is expected though? And maybe change the name of the function to something like wait_for_healthy since 303 is not success, strictly speaking?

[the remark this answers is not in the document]
jeffpw-goog: Here's where that redirect happens: https://github.com/civiform/civiform/blob/a0a95e4ff507081253ddc085e6c9c630df1872cb/server/app/auth/DefaultToGuestRedirector.java

cloud/azure/modules/app/variables.tf
jcma-google: @dkatzz The tier and capacity can be inferred from the sku_name. 

@jeffpw-goog mindlessly copying the old pattern which set defaults inside of variables.tf and overrode them again inside of main.tf. There isn't a good reason for doing this and I think it adds more confusion than not. I'll remove the unnecessary definitions inside of main.tf

cloud/shared/bin/lib/terraform.py
jcma-google: This was leftover from my debugging efforts. Removed it because I have a SKIP_CONFIRMATION variable in my azure config file that adds auto-apply. I've found that setup and deployment fails without auto-apply. Still debugging why that is. I've added a recommendatoin to add this variable to my azure deployment notes.

cloud/azure/bin/lib/key_vault.sh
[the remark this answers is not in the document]
jcma-google: Ah! I didn't realize you had to add echo $? at the end of the expression to get it to work. That's the missing piece, will add back this expression and revert my changes in the other files

cloud/shared/bin/lib/health.sh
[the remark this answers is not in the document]
jcma-google: sg!

cloud/azure/templates/azure_saml_ses/main.tf
jcma-google: done! will test the variable setting once staging is green and if it's not working i'll open a separate issue to track the fix

jcma-google:
@jeffpw-goog @dkatzz hey all, thanks for reviewing this massive cl! and apologies for the size, I had to make a choice between many small cls incrementally progressing towards a functional deployment environment vs. one large cl to get everything working and people can start building features and start testing. All the cls after this will be much smaller!

cloud/azure/bin/lib/key_vault.sh
```diff
@@ -125,4 +125,5 @@ function key_vault::has_secret() {
     -o tsv 2>&1 >/dev/null)"
 
   echo "${SECRET_RESULT}" | grep -q -v "SecretNotFound"
+  echo $
```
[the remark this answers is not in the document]
jeffpw-goog: Related but apparently I forgot to actually post it until just now 🤦 : https://github.com/civiform/cloud-deploy-infra/pull/366/files/d65aabfa07b8158b33e6b25de47504c74b7700a2#r1718788183

cloud/azure/bin/lib/key_vault.sh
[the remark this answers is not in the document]
jcma-google: The reason I changed it is because the current expression doesn't work. The if statement doesn't evaluate as expected, it always returns "true" and never creates the secret even if there is no secret key in the store. I found this out after the script kept on failing on this step and some debug statements pinpointed the error to this step in the process.

No secrets in the keyvault. Script will create since no secrets are present
Screenshot 2024-08-20 at 4 12 32 PM

Script erroneously believes that secrets exist 
Screenshot 2024-08-20 at 4 12 44 PM

This is because the condition `if key_vault::has_secret "${vault}" "${id}"` always evaluates to true. 

When I evaluate the return expression `echo "${SECRET_RESULT}" | grep -q -v "SecretNotFound"` in my terminal I actually get empty string return. 
Screenshot 2024-08-20 at 4 24 09 PM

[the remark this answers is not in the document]
jcma-google: Surprisingly adding echo $ so that the return expression looks like this ` echo "${SECRET_RESULT}" | grep "SecretNotFound" | echo $` does fix things even though it comes with extra new lines  

Screenshot 2024-08-20 at 4 38 06 PM

jcma-google:
@nb1701 (can't respond to your comment about the ses-to-keyvault line for some reason). I'm preserving it for now and following up with another PR (jcma-azure-final-branch-aws) that sets up ses-to-keyvault if the civic entity wants to use AWS as an email service. And I'll change the subprocess back to failing on error.

cloud/azure/bin/deploy
```diff
@@ -22,16 +22,14 @@ fi
...
+if health::wait_for_healthy "${AZURE_PRIMARY_URL}/playIndex"; then
+  echo "New container returns expected signal on ping"
 else
   log::deploy_failed "${IMAGE_TAG}" "${AZURE_USER_ID}"
   azure_log::upload_log_file
   exit 1
 fi
 
-echo "Swapping deployment slot 'canary' into production. This may take up to 10 minutes."
-azure::swap_deployment_slot "${AZURE_RESOURCE_GROUP}" "${AZURE_APP_NAME}"
```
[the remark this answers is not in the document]
jcma-google: I'll be testing the staging to production pipeline next and figuring out if canary is still needed. Will open separate issue to track that.

cloud/azure/templates/azure_saml_ses/bin/setup.py
jcma-google: For some reason github didn't allow me to leave a comment to this reply until now! 

Quoting from my stand-alone response. Preserving the ses-to-keyvault script and gating it behind user input in a separate PR.

cloud/azure/bin/lib/key_vault.sh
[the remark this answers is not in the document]
jcma-google: replied to the comment chain that jeff posted as to why I added the echo $

cloud/azure/bin/lib/key_vault.sh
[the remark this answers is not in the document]
jeffpw-goog: This doesn't work - it just always returns not found. I scheduled some time tomorrow to pair on it :)

cloud/azure/modules/app/variables.tf
```diff
@@ -76,17 +65,18 @@ variable "resource_group_name" {
...
   type        = string
   description = "The sku name for postgres server"
-  default     = "GP_Gen5_2"
+  default     = "GP_Standard_D2s_v3"
 }
 variable "postgres_storage_mb" {
   type        = number
   description = "The mb of storage for postgres instance"
-  default     = 5120
+  default     = 32768
```
gwendolyngoetz, line 79: This is pretty large. After three years in prod Seattle's database is only 275MB out of 5GB available. Our old, pre-terraform install had prod at 50GB. Waste of money. 

Can the storage size be expanded on demand or does it have to be fully allocated upfront when created?

cloud/azure/modules/app/variables.tf
```diff
@@ -76,17 +65,18 @@ variable "resource_group_name" {
 variable "postgres_admin_login" {
   type        = string
   description = "Postgres admin login"
+  default     = "psqladmin"
```
gwendolyngoetz, line 68: Does this need to be a fixed value. The AWS side is generated during the initial setup process?

cloud/azure/bin/lib/key_vault.sh
```diff
@@ -124,5 +124,5 @@ function key_vault::has_secret() {
     --query value \
     -o tsv 2>&1 >/dev/null)"
 
-  echo "${SECRET_RESULT}" | grep -q -v "SecretNotFound"
+  echo "${SECRET_RESULT}" | grep "SecretNotFound" | echo $
```
gwendolyngoetz, line 127: What the intent with this line? It's only ever going to print a "$".

cloud/azure/templates/azure_saml_ses/main.tf
rockycodes: Just came here to see how it went with the server vars! These are indeed safe to be removed and should be removed.

cloud/azure/modules/app/variables.tf
jcma-google: This is the minimum allocation possible for postgres_flexible_database. 
Screenshot 2024-08-21 at 9 18 22 AM

gwendolyngoetz: Huh. The cloud is crazy. Thanks for the info! 🙂

cloud/azure/bin/lib/key_vault.sh
jcma-google: debugging a bash issue. unfortunately the easiest development flow is to commit a change (thus my long list of commits).

cloud/azure/bin/lib/key_vault.sh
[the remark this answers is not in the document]
jcma-google: I think echo $ makes it so that it looks like the vault never has the secret, something about the current expression `echo "${SECRET_RESULT}" | grep -q -v "SecretNotFound"` needs to be fixed. 

I'm going to try your suggestion nick and see if it works.

cloud/azure/modules/app/variables.tf
jcma-google: I had it be fixed because I liked connecting to the DB via azure powershell and not having to look up the username every time. We can change it to be generated following the pattern of {adjective}-{noun} that we use right now for the app service instance but I'm not sure of the value, other than it follows the AWS pattern (and maybe that suffices). 

For now I would like to keep it as psqladmin because changing it would mean modifying this PR further and it's swelled to a large size already. I can add a note to discuss in our next team meeting as to whether the database username should be generated vs. fixed.

dkatzz (approved):
Would be good to create bugs for all the follow ups (if they haven't yet been created), but I'm good with us submitting this (once there's another approver or two) since this gets us a working azure configuration and then we can do separate PRs to add onto it / make updates

cloud/shared/bin/lib/health.sh
```diff
@@ -6,7 +6,7 @@
 #   1: The url for the healthcheck endpoint
 #   2: (Optional) Amount of time to wait before timing out, default
 #######################################
-function health::wait_for_success() {
+function health::wait_for_healthy() {
```
jeffpw-goog, line 9: **todo:** Based on the change you made to use playIndex in cloud/azure/bin/deploy you don't need the changes in this file anymore?

cloud/azure/bin/lib/key_vault.sh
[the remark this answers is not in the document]
jeffpw-goog: Summarizing our offline debug session for posterity: the SECRET_RESULT includes multiple lines, and if the secret is not found only some of those lines include SecretNotFound. So grep -v SecretNotFound always returns true regardless of whether the secret is not found.

Nick's solution works, because it inverts the result of the regular grep instead of using grep -v, but I think this would be more readable:

! echo "${SECRET_RESULT}" | grep -q "SecretNotFound"

cloud/azure/modules/app/variables.tf
jcma-google: yup, can change in a followup PR.

PR #386, opened 2024-09-11
Update azurerm version to 4.1.0 and remove outdated dependencies 

jcma-google:
### Description

- change `retention_duration` to use new name `operational_default_retention_duration` 
- `enforce_private_link_endpoint_network_policies` can be safely removed because private link is not currently set up 
- retention_policy can be safely removed due to [this stackoverflow comment](https://stackoverflow.com/questions/77717130/retention-policy-terraform-is-deprecated-inside-azurerm-monitor-diagnostic-setti) stating that `enabled_log` blocks use the retention policy set in the `azurerm_log_analytics_workspace` resource. 

#### General

- [ x] Added the correct label
- [ x] Created tests which fail without the change (if possible)
- [ x] Performed manual testing (at a minimum run `bin/setup` without your changes and then `bin/deploy` with your changes to ensure your changes don't break existing deployments)
- [x ] Extended the README / documentation, if necessary

### DIFF
--- a/cloud/azure/modules/app/logs.tf
+++ b/cloud/azure/modules/app/logs.tf
@@ -11,67 +11,42 @@ resource "azurerm_monitor_diagnostic_setting" "app_service_log_analytics" {
   target_resource_id         = azurerm_linux_web_app.civiform_app.id
   log_analytics_workspace_id = azurerm_log_analytics_workspace.civiform_logs.id
 
-  log {
+  enabled_log {
     category = "AppServiceAppLogs"
-
-    retention_policy {
-      days    = 0
-      enabled = false
-    }
   }
 
-  log {
+  enabled_log {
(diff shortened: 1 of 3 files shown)


cloud/azure/modules/app/logs.tf
```diff
@@ -11,67 +11,42 @@ resource "azurerm_monitor_diagnostic_setting" "app_service_log_analytics" {
...
+  # }
+  # enabled_log {
+  #   category = "AppServicePlatformLogs"
+  #   enabled  = false
+
+  #   retention_policy {
+  #     days    = 0
+  #     enabled = false
+  #   }
+  # }
```
gwendolyngoetz, line 51: (blocking) If we don't need these go ahead and remove the comments.

PR #387, opened 2024-09-12
Minor v4.1.0 azurerm bugfixes

jcma-google:
### Description

- providers need to include subscription id. This error didn't appear during my testing of the 4.1.0 update because it only occurs when the deployment setup changes.
- change private-ip allocation method to "Static"
  - the staging resource uses a standard SKU and "dynamic" allocation is only permissible for more expensive plans. It looks like "Static" just reserves an IP address and doesn't change it, not that I see a reason to change it after it's allocated. 
- remove `DOCKER_REGISTRY_SERVER_URL` 
  - this is now deprecated. The now preferred way of doing it is to set it in the linux_web_app resource, [which we are already doing](https://github.com/civiform/cloud-deploy-infra/blob/main/cloud/azure/modules/app/main.tf#L76). 
  allocation_method = "Static" 

#### General

- [x ] Added the correct label
- [x ] Created tests which fail without the change (if possible)
- [x ] Performed manual testing (at a minimum run `bin/setup` without your changes and then `bin/deploy` with your changes to ensure your changes don't break existing deployments)
- [ x] Extended the README / documentation, if necessary

### DIFF
--- a/cloud/azure/templates/azure_saml_ses/providers.tf
+++ b/cloud/azure/templates/azure_saml_ses/providers.tf
@@ -1,4 +1,5 @@
 provider "azurerm" {
   features {}
+  # https://github.com/civiform/civiform/issues/8598
   subscription_id = "4ef4ae1b-c966-4ac4-9b7c-a837ea410821"
 }
--- a/cloud/azure/templates/azure_saml_ses/providers.tf
+++ b/cloud/azure/templates/azure_saml_ses/providers.tf
@@ -1,4 +1,4 @@
 provider "azurerm" {
   features {}
-  subscription_id = var.azure_subscription
+  subscription_id = "4ef4ae1b-c966-4ac4-9b7c-a837ea410821"
 }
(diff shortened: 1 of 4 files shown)


cloud/azure/templates/azure_saml_ses/providers.tf
```diff
@@ -1,3 +1,4 @@
 provider "azurerm" {
   features {}
+  subscription_id = "4ef4ae1b-c966-4ac4-9b7c-a837ea410821"
```
gwendolyngoetz, line 3: (question) Is there a reason not to get this from a terraform variable and the civiform*config.sh file. I'm concerned this will get forgotten about and cause issues down the road.

jcma-google: I tried using the config variable here and it works! I didn't realize that the providers.tf could access the variables but on retrospect it makes a lot of sense.

jcma-google: Wait actually it was because I forgot to push my changes

jcma-google: Okay it looks like the providers.tf doesn't have access to the civiform config, but maybe I'm missing something?

gwendolyngoetz (approved):

PR #389, opened 2024-09-13
Add knobs to adjust healcheck parameters

yegle:
### Description

Allow customizing the healthcheck parameters.

Note I added effectively two sets of interval/timeout/retries. It looks like we can't use the same set of values on both ALB and ECS, because ECS allow timeout>interval, while ALB does not allow it.

#### General

- [ ] Added the correct label

Ran `bin/deploy`.

### Issue(s) this completes

Fixes https://github.com/civiform/civiform/issues/7698


cloud/aws/modules/ecs_fargate_service/variables.tf
```diff
@@ -149,3 +149,12 @@ variable "extra_inbound_rule_cidr" {
   type        = string
   default     = null
 }
+
+variable "healthcheck" {
```
dkatzz, line 153: **suggestion:** Thoughts on calling this lb_healthcheck or something to make it a bit more specific?

dkatzz (approved):
One comment, but LGTM! Thank you!

PR #392, opened 2024-09-19
Add private link endpoint to the postgres database

jcma-google:
### Description

Create a private link endpoint and attach it to the postgres database. Add the web app resource to the same virtual network so that it can access the private link. Change the connection string to use the private link name.

Setup and deployed successfully with the changes in this branch.

#### General

- [ x] Added the correct label
- [x ] Created tests which fail without the change (if possible)
- [x ] Performed manual testing (at a minimum run `bin/setup` without your changes and then `bin/deploy` with your changes to ensure your changes don't break existing deployments)
- [x ] Extended the README / documentation, if necessary

### Issue(s) this completes
Fixes #8364


cloud/azure/modules/app/main.tf
```diff
@@ -63,12 +62,18 @@ resource "azurerm_service_plan" "plan" {
...
-  location            = data.azurerm_resource_group.rg.location
-  resource_group_name = data.azurerm_resource_group.rg.name
-  service_plan_id     = azurerm_service_plan.plan.id
-  app_settings        = local.app_settings
+  name                      = "${var.application_name}-${random_pet.server.id}"
+  location                  = data.azurerm_resource_group.rg.location
+  resource_group_name       = data.azurerm_resource_group.rg.name
+  service_plan_id           = azurerm_service_plan.plan.id
+  app_settings              = local.app_settings
+  virtual_network_subnet_id = azurerm_subnet.server_subnet.id
```
dkatzz, line 76: **question:** This was the main change since https://github.com/civiform/cloud-deploy-infra/pull/366 that did it, right?

dkatzz (approved):
Yay!!

cloud/azure/modules/app/main.tf
jcma-google: yep!

PR #397, opened 2024-10-02
Fix restoring from a snapshot

nb1701:
When we restore from a snapshot, Terraform applies the temporary
database password to the restored database, rather than using the actual
database password as applied after the Terraform setup finishes (we do
this so we don't store the actual password in Terraform state). This reads the actual
password for the database from the secret and applies it when restoring.

While the code supports doing a bin/deploy with a snapshot identified
(that is, destroying your current database and restoring from a snapshot
in-place), we've decided to disable this code path for now until we get
a more imperative system for restoring from a snapshot in place in order
to avoid accidental data loss.

In a workflow where you want to set up a separate instance with a
different app_prefix, but based on a snapshot from the old app_prefix,
during bin/setup, this prompts the user to load the secret value from the previous
app_prefix, or to enter the password manually.

Additionally, when we are standing up a brand new CiviForm instance
based on a snapshot from another deployment, that snapshot is encrypted
with the KMS key from the old deployment. This results in Terraform
wanting to destroy the database and recreate it again on the next
deploy. Instead, we now copy the snapshot and re-encrypt it with the new
instance's key, then deploy the new database from that snapshot. A
subsequent deploy with POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER removed will
remove that copy.

Finally, this updates ecs-service-autoscaling to 1.0.10 to squash a bug
that caused it to always add 'null' into alarm action lists.

Fixes https://github.com/civiform/civiform/issues/8085

### DIFF
--- a/cloud/aws/modules/ecs_fargate_service/main.tf
+++ b/cloud/aws/modules/ecs_fargate_service/main.tf
@@ -337,7 +337,7 @@ moved {
 
 module "ecs-autoscaling" {
   source  = "cn-terraform/ecs-service-autoscaling/aws"
-  version = "1.0.9"
+  version = "1.0.10"
 
   name_prefix               = local.name_prefix
   ecs_cluster_name          = var.ecs_cluster_name
--- a/cloud/aws/templates/aws_oidc/bin/aws_cli.py
+++ b/cloud/aws/templates/aws_oidc/bin/aws_cli.py
@@ -4,7 +4,7 @@ import json
 import time
 import inspect
 import re
-from typing import Dict
+from typing import Dict, Optional
 
 from cloud.aws.templates.aws_oidc.bin import resources
 from cloud.shared.bin.lib.config_loader import ConfigLoader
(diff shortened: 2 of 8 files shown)


cloud/aws/modules/ecs_fargate_service/main.tf
```diff
@@ -337,7 +337,7 @@ moved {
 
 module "ecs-autoscaling" {
   source  = "cn-terraform/ecs-service-autoscaling/aws"
-  version = "1.0.9"
+  version = "1.0.10"
```
gwendolyngoetz, line 340: (question) Did renovate miss this? Do you need it for this PR?

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -4,7 +4,7 @@
 import time
 import inspect
 import re
-from typing import Dict
+from typing import Dict, Optional
```
gwendolyngoetz, line 7: (humor) Oh no it's spreading!

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -44,6 +44,41 @@ def update_master_password_in_database(self, db_name: str, password: str):
...
+            self,
+            config: ConfigLoader,
+            user: Optional[str] = None,
+            pwd: Optional[str] = None):
+        """
+        Updates the database to use the given password, and update the secret value,
+        restarting ECS so it can pick it up.
+        """
+        if user is None and pwd is None:
+            raise ValueError("At least one of user or pwd must be provided.")
```
gwendolyngoetz, line 68:
```suggestion
            raise ValueError("Invalid arguents. You must provide a value for 'user', 'password', or both.")
```

cloud/aws/templates/aws_oidc/bin/aws_cli.py
```diff
@@ -44,6 +44,41 @@ def update_master_password_in_database(self, db_name: str, password: str):
...
+            f'{config.app_prefix}-{resources.POSTGRES_PASSWORD}')
+        self.update_master_password_in_database(
+            f'{config.app_prefix}-{resources.DATABASE}', db_password)
+        print('Database password has been set.')
+
+    def set_database_credentials(
+            self,
+            config: ConfigLoader,
+            user: Optional[str] = None,
+            pwd: Optional[str] = None):
```
gwendolyngoetz, line 62: (blocking) When I see pwd my first thought is "present working directory". Recommend making it password for clarity or as least pass.

cloud/aws/templates/aws_oidc/bin/setup.py
```diff
@@ -98,7 +99,29 @@ def post_terraform_setup(self):
         for name, doc in SECRETS.items():
             self._maybe_set_secret_value(
                 f'{self.config.app_prefix}-{name}', doc)
-        self._maybe_change_default_db_password()
+        if self.config.get_config_var('POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER'):
+            print('')
```
gwendolyngoetz, line 103: (question) Is this just for a new line or did you mean to put in text?

cloud/aws/modules/ecs_fargate_service/main.tf
gwendolyngoetz: Gotcha. Just wondering. No need to remove it.

gwendolyngoetz (approved):

PR #403, opened 2024-10-18
Allow an optional cert file to be specified that gets used by pip

dkatzz:
### Description

Charlotte had a new firewall rule installed, which caused an error during deployment: `SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate`.

With this PR, we now allow a cert file to be specified. This also involved a change in their deployment repo, similar to [this](https://github.com/civiform/civiform-deploy/compare/clt-python?expand=1).

Charlotte tested this change and it fixed their issue and allowed them to deploy successfully.

#### General

- [x] Added the correct label

### Issue(s) this completes

https://github.com/civiform/civiform/issues/8943

### DIFF
--- a/cloud/shared/bin/python_env_setup
+++ b/cloud/shared/bin/python_env_setup
@@ -34,7 +34,7 @@ function initialize_python_env() {
   # If a certificate file is specified, perform checks then configure pip to use it
   if [[ -n "$cert_file_path" ]]; then
     # Check if the file exists and is readable
-    if [[ ! -r "$cert_file_path" ]]; then
+    if [[ ! -f "$cert_file_path" || ! -r "$cert_file_path" ]]; then
       echo "Error: Certificate file '$cert_file_path' does not exist or is not readable."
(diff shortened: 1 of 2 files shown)


cloud/shared/bin/python_env_setup
```diff
@@ -30,6 +31,12 @@ function initialize_python_env() {
   # source the activate script (. is more portable between shells than source)
   . .venv/bin/activate
 
+  # If a certificate file is specified, configure pip to use it
+  if [[ -n "$cert_file_path" ]]; then
```
gwendolyngoetz, line 35: (blocking) Same comment as in the other PR. We should test that the path exists and is readable, and of the correct cert format (if possible).

gwendolyngoetz (changes requested):

gwendolyngoetz (approved):

PR #406, opened 2024-10-24
Create a docker image that can be used to deploy CIVIFORM to aws and azure

lagosr-google:
### Description

Create a docker image with the same toolset that exists in the dev container codespace which can be used to deploy CIVIFORM on AWS and AZURE.

The image is published to https://hub.docker.com/r/civiform/deployment-env

New github workflow to build and push the image to docker named "build_push_deployment_env_image" is included.

#### General

- [x] Added the correct label
- [x] Extended the README / documentation, if necessary

Documentation on how to use the docker image to deploy onto azure and in development.


.github/workflows/push_deployment_env_image..yaml
gwendolyngoetz, line 1: (blocking) The file name has two `..` before the extension.

lagosr-google: thank you for spotting that!

PR #419, opened 2024-11-22
Create variable for email provider

dkatzz:
### Description

This will be used so we can conditionally create AWS resources based on this variable (see https://github.com/civiform/cloud-deploy-infra/pull/420)

I think for azure we'll eventually use graph-api as the default, but since that isn't supported, I left it empty for now

#### General

- [x] Added the correct label

### Issue(s) this completes

https://github.com/civiform/civiform/issues/9258

### DIFF
--- a/cloud/azure/templates/azure_saml_ses/variables.tf
+++ b/cloud/azure/templates/azure_saml_ses/variables.tf
@@ -60,6 +60,12 @@ variable "aws_region" {
   default     = "us-east-1"
 }
 
+variable "email_provider" {
+  type        = string
+  description = "The provider to use for sending emails"
+  default     = "aws_ses"
+}
+
 variable "sender_email_address" {
   type        = string
   description = "Email address that emails will be sent from"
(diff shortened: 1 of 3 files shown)


cloud/azure/templates/azure_saml_ses/variables.tf
```diff
@@ -60,6 +60,11 @@ variable "aws_region" {
   default     = "us-east-1"
 }
 
+variable "email_provider" {
+  type        = string
+  description = "The provider to use for sending emails"
+}
```
rockycodes, line 66: question (non-blocking): why do you need a default for aws but not azure?

rockycodes (approved):

cloud/azure/templates/azure_saml_ses/variables.tf
dkatzz: Good question. I think for azure we'll eventually use graph-api as the default, but since that isn't supported I left it empty for now. I'll add it to the description.

PR #440, opened 2025-02-04
This fixes an issue with the listener not being set up correctly

dkatzz:
### Description

Listener should only be created after the security group rule is

#### General

- [ ] Added the correct label

### Issue(s) this completes

### DIFF
--- a/cloud/aws/modules/ecs_fargate_service/main.tf
+++ b/cloud/aws/modules/ecs_fargate_service/main.tf
@@ -211,6 +211,10 @@ resource "aws_lb_listener" "lb_http_listeners" {
   }
 
   tags = var.tags
+
+  depends_on = [
+    aws_security_group_rule.ingress_through_http[0]
+  ]
 }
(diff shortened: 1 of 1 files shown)


cloud/aws/modules/ecs_fargate_service/main.tf
```diff
@@ -211,6 +211,10 @@ resource "aws_lb_listener" "lb_http_listeners" {
   }
 
   tags = var.tags
+
+  depends_on = [
+    aws_security_group_rule.ingress_through_http[0]
+  ]
```
gwendolyngoetz, line 217: Does depends on require at least one to be present or will it work if there are no listeners?

dkatzz: I tried with enable_http set to false and it still works because it won't try to create this listener if that is set to false

gwendolyngoetz (approved):

PR #442, opened 2025-02-12
Remove reliance on AWS from the azure deployment

dkatzz:
### Description

We previously relied on AWS from the Azure deployment for email sending, but we're replacing that with Graph API, since that is what Miami Dade would like to use and they're currently the only deployment on azure. In the future, we can add back AWS support if needed, but conditionally creating resources is not straightforward, so it didn't seem worth investing in that right now. Got sign off on this from the eng team in our weekly meeting.

#### General

- [ ] Added the correct label

### Issue(s) this completes


rockycodes (approved):
This looks good to me, but I don't have a lot of context on how things work on the Azure side so I defer to you and Ricardo for that. Is there any documentation that needs to be updated?

PR #450, opened 2025-03-10
Change default LB policy

nb1701:
This sets the default Load Balancer policy to
ELBSecurityPolicy-TLS13-1-2-2021-06, which allows TLSv1.3 and 1.2. This is the recommended policy. See
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/describe-ssl-policies.html for details on all the policies.

In fact, clients were still able to connect via TLSv1.3 due to changes in the ALB applied last year. https://aws.amazon.com/about-aws/whats-new/2023/03/application-load-balancer-tls-1-3/

This policy ensures that connections between the ALB and the server is also TLSv1.3 or 1.2.  This is somewhat less important since it is all internal to the VPC, but a good idea nonetheless.

This also updates the ALB policy when setting up pgadmin.

Finally, this exposes the security policy as a variable so it can be overriden if needed.

#### General

- [x] Added the correct label

Once deployed, do `curl -vv https://<your instance>/` to ensure it is connect with TLSv1.3 (although that should happen without this change too).

### Issue(s) this completes

https://github.com/civiform/civiform/issues/6730

### DIFF
--- a/cloud/aws/bin/pgadmin.py
+++ b/cloud/aws/bin/pgadmin.py
@@ -27,7 +27,7 @@ def run(config: ConfigLoader):
 
     url = f"{config.get_base_url()}:4433"
     print(
-        "\npgadmin terraform deployment finished. Waiting for pgadmin to be available (some request failures are expected). Press ctlr-c to shortcut this wait and print connection information."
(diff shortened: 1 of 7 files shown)


dkatzz (approved):
LGTM - let's just make sure some info on this gets into the release notes, so people know to expect the update

PR #461, opened 2025-04-04
Create a helper function to get the formatted name for bastion

dkatzz:
### Description

Still getting the error:

│ Error: unable to assume default computer name "computer_name" cannot contain the special characters: `\/"[]:|<>+=;,?*@&~!#$%^()_{}'` Please adjust the "name", or specify an explicit "computer_name"

Realized this was another place we set the bastion name.

#### General

- [ ] Added the correct label

### Issue(s) this completes

### DIFF
--- a/cloud/azure/bin/lib/bastion.sh
+++ b/cloud/azure/bin/lib/bastion.sh
@@ -7,7 +7,7 @@
 #   1: the resource group name
 #######################################
 function get_formatted_name() {
-  echo "$(echo "${1:0:15}" | sed 's/[^[:alnum:]]//g')"
+  echo "${1:0:15}" | sed 's/[^[:alnum:]]//g'
 }
 
 #######################################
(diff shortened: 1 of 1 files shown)


cloud/azure/bin/lib/bastion.sh
```diff
@@ -1,5 +1,15 @@
 #! /usr/bin/env bash
 
+#######################################
+# Helper function to get the formatted name of the
+# resource group, which is alphanumeric only
+# Arguments:
+#   1: the resource group name
+#######################################
+function get_formatted_name() {
+  echo "$(echo "${1:0:15}" | sed 's/[^[:alnum:]]//g')"
```
gwendolyngoetz, line 10:
```suggestion
  echo "${1:0:15}" | sed 's/[^[:alnum:]]//g'
```
Double check, but i think this will still work.

dkatzz: Thanks! That seems to work!

gwendolyngoetz (approved):

PR #468, opened 2025-05-07
Enable restoring from a snapshot on deploy

rockycodes:
### Description

This PR re-enables allowing deployments to restore the db to a db snapshot on deploy (not just on setup). It adds a user input check to help guard against data loss from accidentally deploying with this value set.

#### General

- [x] Added the correct label
- [ ] Extended the README / documentation, if necessary -- will do this soon! maybe [here](https://docs.civiform.us/it-manual/sre-playbook/disaster-recovery)

1. Spin up a deployment and seed the database. Publish all programs.
2. Take a snapshot of the db in AWS
3. Clear the database via the developer tools
4. Set `export POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER=<snapshot name>` in your deployment config
5. Run `bin/deploy` and follow the prompts
6. Deployment should succeed and you should see data re-populated on your site again

### Issue(s) this completes

Slack thread with context: https://civiform.slack.com/archives/C044VUMJSSH/p1746624121481959

### DIFF
--- a/cloud/aws/bin/deploy.py
+++ b/cloud/aws/bin/deploy.py
@@ -16,16 +16,21 @@ def run(config: ConfigLoader):
         _check_application_secret_length(config, aws)
         _check_for_postgres_upgrade(config, aws)
 
-    # Even though the code supports it, for now, we are disallowing doing bin/deploy
-    # with a snapshot identified in order to prevent accidental data loss. When we move to
-    # a more imperative process for restoring from a snapshot, we can remove this block.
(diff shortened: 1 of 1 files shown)


dkatzz (approved):
Curious if data loss happened through this before or if we're just worried that someone sets this and forgets and then loses data between deployments? Once we get that history, maybe we decide to only allow this for non-prod or do some stricter check, but I think the warning makes sense for now.

PR #469, opened 2025-05-08
Create a variable for CIDR block for sg ingress

dkatzz:
### Description

There was a request to be able to lock down staging sites to a particular CIDR range

#### General

- [ ] Added the correct label

### Issue(s) this completes

### DIFF
--- a/cloud/aws/templates/aws_oidc/variable_definitions.json
+++ b/cloud/aws/templates/aws_oidc/variable_definitions.json
@@ -449,6 +449,12 @@
     "tfvar": true,
     "type": "string"
   },
+  "INGRESS_SG_CIDR": {
+    "required": false,
+    "secret": false,
+    "tfvar": true,
+    "type": "string"
+  },
   "ENABLE_HTTP_LISTENER": {
     "required": false,
     "secret": false,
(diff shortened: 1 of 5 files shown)


rockycodes (approved):

cloud/aws/templates/aws_oidc/variable_definitions.json
```diff
@@ -449,6 +449,12 @@
     "tfvar": true,
     "type": "string"
   },
+  "INGRESS_SG_CIDR": {
```
rockycodes, line 452: question (non-blocking): Is there an order to how variables are entered in here?

dkatzz: not really - I put it next to something somewhat similar but if there's a different order you think would be better, I can update it!

rockycodes: nope, i think this is fine. but might be nice to alphabetize the whole list in a separate PR so it's easier to find/place things. or organize it by category. what do you think?

rockycodes: Great idea, I'll ask.

PR #470, opened 2025-05-08
Turn off delete_automated_backups

rockycodes:
### Description

This PR adds a configurable variable `delete_automated_backups` that controls whether or not system/auto-generated snapshots will be deleted when the associated db is destroyed. We default this to `false` to override Terraform's default which automatically deletes system snapshots when the associated db is torn down. We allow this variable to be configurable so admin can allow system snapshots to be deleted if they want to.

We decided to make this change after attempting to restore a gov environment to a system snapshot only to have it deleted in the process.

Note: Manual snapshots are not deleted when the db is destroyed and are not affected by this change.

#### General

- [x] Added the correct label
- [ ] Extended the README / documentation, if necessary (will update when I do the other docs updates)

I tested this in the following ways:
- `bin/setup` with neither `POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER` nor `DELETE_AUTOMATED_BACKUPS` (no effect)
- `bin/deploy` with neither `POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER` nor `DELETE_AUTOMATED_BACKUPS` (no effect)
- `bin/deploy` with `POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER` set to a snapshot and `DELETE_AUTOMATED_BACKUPS` not set (retains system snapshots)
- `bin/deploy` with `POSTGRES_RESTORE_SNAPSHOT_IDENTIFIER` set to a manual snapshot and `DELETE_AUTOMATED_BACKUPS` set to true (previous system snapshots are maintained, new ones are deleted if/when the db is torn down again)

### DIFF
--- a/cloud/aws/templates/aws_oidc/main.tf
+++ b/cloud/aws/templates/aws_oidc/main.tf
@@ -61,7 +61,7 @@ resource "aws_db_instance" "civiform" {
 
   # If not null, destroys the current database, replacing it with a new one restored from the provided snapshot
   snapshot_identifier             = var.postgres_restore_snapshot_identifier == null ? null : aws_db_snapshot_copy.copied_snapshot[0].target_db_snapshot_identifier
-  delete_automated_backups        = false
+  delete_automated_backups        = var.delete_automated_backups
   deletion_protection             = local.deletion_protection
   instance_class                  = var.postgres_instance_class
   allocated_storage               = var.postgres_storage_gb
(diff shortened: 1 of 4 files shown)


cloud/aws/templates/aws_oidc/main.tf
```diff
@@ -61,6 +61,7 @@ resource "aws_db_instance" "civiform" {
 
   # If not null, destroys the current database, replacing it with a new one restored from the provided snapshot
   snapshot_identifier             = var.postgres_restore_snapshot_identifier == null ? null : aws_db_snapshot_copy.copied_snapshot[0].target_db_snapshot_identifier
+  delete_automated_backups        = var.delete_automated_backups
```
dkatzz, line 64: **suggestion:** we could consider renaming to delete_automated_db_backups or delete_automated_rds_backups to make it clear that this is for the database in particular. Thoughts?

dkatzz (approved):
One thought on naming but otherwise LGTM!

cloud/aws/templates/aws_oidc/main.tf
rockycodes: i like it! updated.

dkatzz (approved):

PR #475, opened 2025-05-19
Revert "Fix latest SHA lookup (#473)"

rockycodes:
This reverts commit 1102b9d378dbf4dd3fb03d65a5a93333a0896faa.

We are seeing a new issue when deploying with https://github.com/civiform/cloud-deploy-infra/commit/1102b9d378dbf4dd3fb03d65a5a93333a0896faa, so reverting for now which we think will fix it.


rockycodes:
jk Danielle already had a PR out for this!

